PR #1, opened 2019-10-01
Add App Mesh controller and injector charts

stefanprodan:
This PR adds the App Mesh controller and injector charts and CircleCI config (lint and validate charts).

Tested on EKS v1.13 with Helm v2 and v3:

```sh
kubectl create ns appmesh-system

kubectl apply -f stable/appmesh-controller/crds/

helmv3 upgrade -i appmesh-controller stable/appmesh-controller/ \
--wait --namespace appmesh-system

helmv3 upgrade -i appmesh-inject stable/appmesh-inject/  \
--wait --namespace appmesh-system \
--set mesh.create=true \
--set mesh.name=global \
--set mesh.discovery=dns

kubectl get meshes.appmesh.k8s.aws -oyaml
  status:
    meshCondition:
    - lastTransitionTime: "2019-10-01T22:07:50Z"
      status: "True"
      type: MeshActive

kubectl -n appmesh-system logs deploy/appmesh-controller
I1001 22:07:34.124266       1 root.go:123] Running controller with threadiness=5
I1001 22:07:34.124281       1 controller.go:180] Starting controller
I1001 22:07:34.124287       1 controller.go:190] Waiting for informer caches to sync
I1001 22:07:34.224579       1 controller.go:195] Starting workers
I1001 22:07:34.224755       1 controller.go:205] Started workers
I1001 22:07:50.410786       1 mesh.go:58] Created mesh global

kubectl -n appmesh-system logs deploy/appmesh-inject
I1002 07:46:07.889972       1 server.go:74] Starting HTTP server on port 8080
I1002 07:51:42.813420       1 server.go:238] Patching pod { podinfo-7fcd7db55-     0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[app:podinfo pod-template-hash:7fcd7db55] map[kubernetes.io/psp:eks.privileged prometheus.io/scrape:true] [{apps/v1 ReplicaSet podinfo-7fcd7db55 7974861e-e4e9-11e9-9771-0657df3da4b0 0xc0004977e7 0xc0004977e8}] nil []  []}
```

Fix: #3 
Fix: #2


stefanprodan:
@nckturner this is ready for review

README.md
```diff
@@ -1,13 +1,41 @@
...
+```
+
+Install the App Mesh admission controller:
+
+```sh
+helm upgrade -i appmesh-inject eks/appmesh-inject \
+--namespace appmesh-system \
+--set mesh.create=true \
+--set mesh.name=global
+```
```
nckturner: Below this block, maybe we can have a note about upgrading the existing installation before the helm chart existed.

stable/appmesh-controller/crds/crds.yaml
```diff
@@ -0,0 +1,382 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: meshes.appmesh.k8s.aws
```
nckturner: It would be nice if there was a way to not have to define these both in the controller package and the helm chart, but I feel like we should keep them in both places for now, for non helm users.

stable/appmesh-controller/templates/_helpers.tpl
```diff
@@ -0,0 +1,56 @@
...
+Common labels
+*/}}
+{{- define "appmesh-controller.labels" -}}
+app.kubernetes.io/name: {{ include "appmesh-controller.name" . }}
+helm.sh/chart: {{ include "appmesh-controller.chart" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
```
nckturner: According to the helm docs: 

> Release.Service: The name of the releasing service (always Tiller).

Is it different in helm V3?  Should we only set this if `.Release.Service` itself is set?

stable/appmesh-controller/templates/_helpers.tpl
```diff
@@ -0,0 +1,56 @@
...
+{{- end -}}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "appmesh-controller.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
+    {{ default (include "appmesh-controller.fullname" .) .Values.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.serviceAccount.name }}
```
nckturner: Does this mean that if `.Values.serviceAccount.create` is specified, then we fallback on the `appmesh-controller.fullname`, but otherwise we fallback on the default service account?  So `.Values.serviceAccount.name` can either refer to a service account that helm creates, or an existing service account, correct?

README.md
```diff
@@ -1,13 +1,41 @@
...
+Apply the App Mesh CRDs:
+
+```sh
+kubectl apply -f https://raw.githubusercontent.com/aws/eks-charts/master/stable/appmesh-controller/crds/crds.yaml
+```
+
+Install the App Mesh CRD controller:
+
+```sh
+helm upgrade -i appmesh-controller eks/appmesh-controller --namespace appmesh-system
```
nckturner: So is namespace expected to always be passed?  Can we add a default namespace?

stefanprodan: The default namespace is `default` same as with kubectl. Helm v3 doesn't support namespace definitions and that's a good thing :) it uses the kubectl context for that.

stable/appmesh-controller/templates/_helpers.tpl
stefanprodan: Yes, this allows a user to specify an existing service account and combined with `rbac.create=false` you can have full control over what the App Mesh controllers can do.

stable/appmesh-controller/templates/_helpers.tpl
stefanprodan: I've used the Helm v3 template when creating these charts, the `.Release.Service` is handled by v3 like this `app.kubernetes.io/managed-by: Helm`.

stable/appmesh-controller/crds/crds.yaml
stefanprodan: The CRDs in the controller repo should be for development purpose only. The App Mesh users should use the CRDs from this repo as it doesn't require Helm to install them.

README.md
stefanprodan: Done!

PR #16, opened 2019-10-17
App Mesh Prometheus: exclude high-cardinality cAdvisor metrics

stefanprodan:
This PR prevents Prometheus OOM on large clusters by excluding high-cardinality cAdvisor metrics from App Mesh Prometheus database.

### DIFF
--- a/stable/appmesh-prometheus/templates/config.yaml
+++ b/stable/appmesh-prometheus/templates/config.yaml
@@ -74,35 +74,19 @@ data:
       scheme: https
       tls_config:
         ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+        insecure_skip_verify: true
       bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
       relabel_configs:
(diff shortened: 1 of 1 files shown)


stable/appmesh-prometheus/templates/config.yaml
```diff
@@ -116,7 +100,16 @@ data:
...
         replacement: /api/v1/nodes/${1}/proxy/metrics/cadvisor
 
-    # scrape config for pods
+      # Exclude high cardinality metrics
+      metric_relabel_configs:
+      - source_labels: [__name__]
+        regex: (container|machine)_(cpu|memory|network|fs)_(.+)
+        action: keep
+      - source_labels: [__name__]
+        regex: container_memory_failures_total
```
nckturner: Why is this any different than other container metrics?

stefanprodan: Dropping that metric helps reduce the disk space usage dramatically. It a highly cardinality metric that we will not need in  any App Mesh dashboards.

PR #24, opened 2019-10-23
add persistent storage example

masterkain:
`retention` needs a mention in the table too?


stefanprodan:
Yes I forgot about it, please add it to the table too. Thanks

stable/appmesh-prometheus/README.md
```diff
@@ -56,4 +82,5 @@ Parameter | Description | Default
 `rbac.pspEnabled` | If `true`, create and use a restricted pod security policy | `false`
 `serviceAccount.create` | If `true`, create a new service account | `true`
 `serviceAccount.name` | Service account to be used | None
+`retention` |  When to remove old data | `15d`
```
stefanprodan: The default retention is 6h see https://github.com/aws/eks-charts/blob/master/stable/appmesh-prometheus/values.yaml#L5 Also the `scrapeInterval: 5s` should be added.

stefanprodan:
@nckturner this is ready to be merged

PR #27, opened 2019-10-29
Adding chart for the AWS CNI

max-rocket-internet:
This is essentially a copy of [v1.5.3](https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/v1.5.3/config/v1.5/aws-k8s-cni.yaml) release YAML but with the standard Helm labels and options added.

I tested it and it works for me. I did a YAML diff of a pod and the daemonset between what this chart produces and what's in an EKS cluster by default and there's no major differences apart from labels and `securityContext.procMount: Default` but this isn't in the release YAML so I guess it's not required.

Some notes....

1. **Naming**: `amazon-vpc-cni-k8s` is not a good name for a chart and neither is `aws-node` so I've chosen `aws-vpc-cni` and set the default name override options to `aws-node` so as to be compatible with current EKS design.

2. **Installation**:  To install it over the top of the existing `aws-node` resources is super easy. Helm just overwrites them. See the README 😃

3. **Resource labels**: These are the default Helm labels for any change but I also added `k8s-app: aws-node` for backwards compatibility.

Related issues:
https://github.com/aws/amazon-vpc-cni-k8s/issues/336
https://github.com/aws/amazon-vpc-cni-k8s/issues/679

### DIFF
--- a/stable/aws-vpc-cni/Chart.yaml
+++ b/stable/aws-vpc-cni/Chart.yaml
@@ -17,7 +17,7 @@ maintainers:
   - name: Nicholas Turner
     url: https://github.com/nckturner
     email: nckturner@users.noreply.github.com
-  - name: Stefan Prodan
-    url: https://github.com/stefanprodan
-    email: stefanprodan@users.noreply.github.com
+  - name: Claes Mogren
+    url: https://github.com/mogren
+    email: mogren@users.noreply.github.com
 engine: gotpl
--- a/stable/aws-vpc-cni/Chart.yaml
+++ b/stable/aws-vpc-cni/Chart.yaml
@@ -1,0 +1,23 @@
+apiVersion: v1
+name: aws-vpc-cni
+version: 1.0.0
+appVersion: "v1.5.3"
+description: A Helm chart for the AWS VPC CNI
+home: https://github.com/aws/amazon-vpc-cni-k8s
+sources:
+  - https://github.com/aws/amazon-vpc-cni-k8s
+description: Networking plugin repository for pod networking in Kubernetes using Elastic Network Interfaces on AWS
+icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
+keywords:
+  - eks
+  - cni
+  - networking
+  - vpc
+maintainers:
+  - name: Nicholas Turner
+    url: https://github.com/nckturner
+    email: nckturner@users.noreply.github.com
+  - name: Stefan Prodan
+    url: https://github.com/stefanprodan
+    email: stefanprodan@users.noreply.github.com
+engine: gotpl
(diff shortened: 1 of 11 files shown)


max-rocket-internet:
> Let's merge this even if Helm v3 tests are failing, I'll find a solution to make the chart compatible with both v2 and v3.

Helm 3 does not even have a stable release yet. Sounds good.

stable/aws-vpc-cni/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+keywords:
+  - eks
+  - cni
+  - networking
+  - vpc
+maintainers:
+  - name: Nicholas Turner
+    url: https://github.com/nckturner
+    email: nckturner@users.noreply.github.com
+  - name: Stefan Prodan
```
nckturner: Can you swap Stefan with Claes Mogren?

stable/aws-vpc-cni/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Installing the Chart
+
+First add the EKS repository to Helm:
+
+```shell
+helm repo add eks https://aws.github.io/eks-charts
+```
+
+To install the chart with the release name `aws-node` and default configuration:
```
nckturner: Should we use `aws-vpc-cni` as the chart name/install name as well?  As long as the daemonset is called aws-node everything should be ok... @mogren thoughts?

mogren: I agree that `aws-vpc-cni` is a better name, but I think people are too used to `aws-node` by now. Would it be confusing if the daemonset has a different name?

stable/aws-vpc-cni/Chart.yaml
max-rocket-internet: Sure!

stable/aws-vpc-cni/README.md
max-rocket-internet: This is just the release name. I chose `aws-node` as an example in the README but it can be anything and with the default configuration the resources will be named `aws-node*` regardless due to `fullnameOverride` and `nameOverride` in `values.yaml`.

It's unconventional to set `fullnameOverride` and `nameOverride` by default but I think it makes sense given the current situation.

Other option is we rename the whole chart to `aws-node` and leave the name override options blank as is convention. But `aws-node` is a not a great name...

Up to you how you want it 🙂

jaypipes: Let's call the chart `aws-vpc-cni` please. As @nckturner mentioned, as long as the Daemonset remains `aws-node`, we're good.

max-rocket-internet: It is named that: https://github.com/aws/eks-charts/pull/27/files#diff-d8c9f0ffa17403764b45ca33421e5d4dR2

jaypipes: Sorry, I meant the release name, not the chart name.

max-rocket-internet: OK cool. Done! To be clear, this is cosmetic though. Users can choose their own name for the release.

max-rocket-internet:
Can we merge this now?

jaypipes:
@max-rocket-internet the helm3 issue (#32) merged. I'm good merging this now.

PR #37, opened 2019-11-21
Adding aws-node-termination-handler-chart

jillmon:
*Issue #:* https://github.com/aws/aws-node-termination-handler/issues/7

*Description of changes:*
Adding helm chart for the aws-node-termination-handler located at https://github.com/aws/aws-node-termination-handler

### DIFF
--- a/stable/aws-node-termination-handler-chart/README.md
+++ b/stable/aws-node-termination-handler-chart/README.md
@@ -1,0 +1,34 @@
+# AWS Node Termination Handler Chart
+AWS Node Termination Handler Helm chart for Kubernetes. For more information on this project see the project repo at https://github.com/aws/aws-node-termination-handler. 
+## Prerequisite
+* Kubernetes >= 1.11
+## Installing the Chart
+Add the EKS repository to Helm:
+```sh
+helm repo add eks https://aws.github.io/eks-charts
+```
+Install AWS Node Termination Handler:
+```sh
+helm upgrade -i aws-node-termination-handler eks/aws-node-termination-handler
+```
+## Uninstalling the Chart
+
+To uninstall/delete the `aws-node-termination-handler` deployment:
+
+```console
+helm delete --purge aws-node-termination-handler
+```
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+The following tables lists the configurable parameters of the chart and their default values.
+| Parameter           | Description  | Default |
+|---------------------|--------------|---------|
+| `name`              | Main project keyword. All pods, cluster roles, service accounts, etc. are created under this name. | `"node-termination-handler"` |
+| `namespace`         | The [kubernetes namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/) | `"kube-system"` |
+| `imageName`         | Refers to docker image located [here](https://hub.docker.com/r/amazon/aws-node-termination-handler). | `"amazon/aws-node-termination-handler"` |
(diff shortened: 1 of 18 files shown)


max-rocket-internet:
Awesome!

If we wanted to make this chart match the standard Helm conventions then there would be a few changes required. For example there are many missing missing labels and `.Values.name` is not the conventional way of changing the name of the resources. If you want to see what mean, you can run `helm create test-chart` and compare the output (this creates a skeleton chart following the current conventions) 🙂

stable/aws-node-termination-handler-chart/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+Install AWS Node Termination Handler:
+```sh
+helm upgrade -i aws-node-termination-handler eks/aws-node-termination-handler
+```
+## Uninstalling the Chart
+
+To uninstall/delete the `aws-node-termination-handler` deployment:
+
+```console
```
jaypipes: consider being consistent with the blocks above and use `sh` instead of `console` here.

stable/aws-node-termination-handler-chart/README.md
```diff
@@ -0,0 +1,34 @@
...
+| `namespace`         | The [kubernetes namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/) | `"kube-system"` |
+| `imageName`         | Refers to docker image located [here](https://hub.docker.com/r/amazon/aws-node-termination-handler). | `"amazon/aws-node-termination-handler"` |
+| `imageVersion`      | Refers to current docker image version found [here](https://hub.docker.com/r/amazon/aws-node-termination-handler/tags). | `"v1.0.0"` |
+| `deleteLocalData`   | Tells kubectl to continue even if there are pods using emptyDir (local data that will be deleted when the node is drained). | `"false"` |
+| `ignoreDaemonsSets` | Causes kubectl to skip daemon set managed pods | `"true"` |
+| `gracePeriod`       | The time in seconds given to each pod to terminate gracefully. If negative, the default value specified in the pod will be used. | `"30"` |
+| `nodeSelector`      | Tells the daemon set where to place the node-termination-handler pods. For example, setting this to `lifecycle: "Ec2Spot"` would only place pods on nodes with that label. | `{}` |
```
jaypipes: Using `on-demand: "false"` as the example would be more in-line with kops-based example setups. Or possibly something like `aws.amazon.com/purchaseType: "spot"`

stable/aws-node-termination-handler-chart/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+namespace: "kube-system"
+
+# image values
+imageName: "amazon/aws-node-termination-handler"
+imageVersion: "v1.0.0"
+
+# deleteLocalData tells kubectl to continue even if there are pods using
+# emptyDir (local data that will be deleted when the node is drained).
+deleteLocalData: \"false\"
```
jaypipes: here and below for ignoreDaemonSets and gracePeriod, you don't need the backslash escapes.

stable/aws-node-termination-handler-chart/README.md
jaypipes: You might also want to mention that this value needs to be a valid YAML expression.

max-rocket-internet: I think `lifecycle` is the correct term here, no?

From the AWS console:
Screen Shot 2019-11-22 at 16 03 13

jaypipes: Typically I see people refer to instances as either "on-demand" or "spot" and refer to the distinction as "purchase type". But, who knows? :)

PR #38, opened 2019-11-25
Updated markdown table to render properly, deleted duplicated code

mattrandallbecker:
*Description of changes:*
Looks like Jillian's chart PR accidentally included a duplicated directory and her markdown table rendering didn't work. Deleted duplicated code and fixed the table.


stefanprodan:
I think we need to add labels, psp, rbac and metadata to this chart the same way all the other have. Also the namespace and overrides shouldn't be hardcoded. @nckturner I would hold back on releasing until of this gets fixed.

PR #28, opened 2019-10-29
Adding chart for AWS' new EKS Pod Identity Webhook

max-rocket-internet:
Note the chart won't run yet as we are still waiting on official docker images: https://github.com/aws/amazon-eks-pod-identity-webhook/issues/5

Related issues:
https://github.com/aws/amazon-eks-pod-identity-webhook/issues/4
https://github.com/aws/containers-roadmap/issues/23

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #48, opened 2019-12-11
Adding annotation to service account

mohsen0:
In order to be able to assign the related role 
https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html

### DIFF
--- a/stable/aws-vpc-cni/Chart.yaml
+++ b/stable/aws-vpc-cni/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: aws-vpc-cni
-version: 1.0.1
+version: 1.0.2
 appVersion: "v1.5.5"
 description: A Helm chart for the AWS VPC CNI
 home: https://github.com/aws/amazon-vpc-cni-k8s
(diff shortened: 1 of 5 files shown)


mohsen0:
I updated the readme, But I am not sure if I could add the related env vars directly to the daemonset, since without the EKS Pod Identity Webhook this should work too.
and EKS Pod Identity Webhook chart is not ready yet, https://github.com/aws/eks-charts/pull/28

mohsen0:
I got this working for Kubernetes 1.14 in EKS, So Documentation is very confusing, I might raise some changes for docs as well. EKS Pod Identity Webhook already deployed in the background since, after annotation of the service account with the role, the env variables and volume for the aws-iam-token are added to the pod.
So I don't think we need to have https://github.com/aws/eks-charts/pull/28
it seems it is running next to k8s control plan

```yaml
$ kubectl get mutatingwebhookconfigurations.admissionregistration.k8s.io pod-identity-webhook --output=yaml
apiVersion: admissionregistration.k8s.io/v1beta1
kind: MutatingWebhookConfiguration
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: | ...
  creationTimestamp: "2019-11-25T17:13:29Z"
  generation: 1
  name: pod-identity-webhook
  resourceVersion: ...
  selfLink: /apis/admissionregistration.k8s.io/v1beta1/mutatingwebhookconfigurations/pod-identity-webhook
  uid: ...
webhooks:
- admissionReviewVersions:
  - v1beta1
  clientConfig:
    caBundle:  ...
    url: https://127.0.0.1:23443/mutate
```
But one thing that pod-identity-webhook doesnt do is mounting of the service account token volume which by adding `automountServiceAccountToken: true` should be done automatically.

This PR is ready for review.

mohsen0:
@jaypipes Can you have a look at this PR again? it is really a simple change

PR #64, opened 2020-01-30
Add option to set up pod annotations for appmesh controller

pipo02mix:
Issue #, if available:
No issue

Description of changes:

Add option in appmesh controller chart to be able to define pod annotations

Signed-off-by: pipo02mix <fernando@giantswarm.io>

### DIFF
--- a/stable/appmesh-controller/values.yaml
+++ b/stable/appmesh-controller/values.yaml
@@ -38,3 +38,6 @@ rbac:
   create: true
   # rbac.pspEnabled: `true` if PodSecurityPolicy resources should be created
   pspEnabled: false
+
+# kiam-role name in case you want to annotate the pods to use kiam to get the AWS role
+kiam-role: ""
--- a/stable/appmesh-controller/values.yaml
+++ b/stable/appmesh-controller/values.yaml
@@ -27,6 +27,8 @@ tolerations: []
 
 affinity: {}
 
+podAnnotations: {}
+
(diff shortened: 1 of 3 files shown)


mohsen0:
The pattern for other charts are different, there is the ‘serviceAccount.annotations’

mohsen0:
https://github.com/aws/eks-charts/blob/master/stable/aws-vpc-cni/values.yaml#L41

stable/appmesh-controller/values.yaml
```diff
@@ -38,3 +38,6 @@ rbac:
   create: true
   # rbac.pspEnabled: `true` if PodSecurityPolicy resources should be created
   pspEnabled: false
+
+# kiam-role name in case you want to annotate the pods to use kiam to get the AWS role
+kiam-role: ""
```
stefanprodan: Can you please change this to `kiamRole` and remove index from the if condition?

pipo02mix:
ok I saw you already have something similar in vpc-cni

https://github.com/aws/eks-charts/blob/master/stable/aws-vpc-cni/values.yaml#L29

So I have used same approach, is that ok?

stefanprodan:
@pipo02mix please change the commit message, PR name and description to reflect the current changes.

pipo02mix:
done

pipo02mix:
How/when this chart is released? 🤔

PR #69, opened 2020-02-06
Allow users to set log verbosity (level) for appmesh-controller

kiranmeduri:
Issue #, if available:

Description of changes:

Tested with debug and info level
```
helm upgrade -i appmesh-controller ./stable/appmesh-controller \
--namespace appmesh-system \
--set logging.verbosity=4
```

### DIFF
--- a/stable/appmesh-controller/templates/deployment.yaml
+++ b/stable/appmesh-controller/templates/deployment.yaml
@@ -32,8 +32,10 @@ spec:
             {{- if .Values.region }}
             - --aws-region={{ .Values.region }}
             {{- end }}
-            {{- if .Values.logging.verbosity }}
-            - --v={{ .Values.logging.verbosity  }}
+            {{- if eq .Values.log.level "debug" }}
+            - --v=4
+            {{- else if eq .Values.log.level "info" }}
+            - --v=0
             {{- end }}
           livenessProbe:
             exec:
--- a/stable/appmesh-controller/templates/deployment.yaml
+++ b/stable/appmesh-controller/templates/deployment.yaml
@@ -32,6 +32,9 @@ spec:
             {{- if .Values.region }}
             - --aws-region={{ .Values.region }}
             {{- end }}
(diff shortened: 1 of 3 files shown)


stable/appmesh-controller/values.yaml
```diff
@@ -40,3 +40,7 @@ rbac:
   create: true
   # rbac.pspEnabled: `true` if PodSecurityPolicy resources should be created
   pspEnabled: false
+
+logging:
+  #logging.verbosity: <=3 (INFO), >=4 (DEBUG)
+  verbosity: 0
```
bcelenza: Are the verbosity numbers actually related to string level descriptions (i.e. 3 is INFO, 4 is DEBUG), or are these just loose translations? If the former, it might be clearer to allow setting the log level by the string description.

kiranmeduri: This is because of v-style logging in Kubernetes ecosystem. See https://godoc.org/github.com/kubernetes/klog. I can change it but that means we need to be consistent in using V(n) in log statements.

kiranmeduri: I will make this change as users will not know what number to set. Thanks

stable/appmesh-controller/templates/deployment.yaml
```diff
@@ -32,6 +32,11 @@ spec:
             {{- if .Values.region }}
             - --aws-region={{ .Values.region }}
             {{- end }}
+            {{- if eq .Values.log.level "debug" }}
+            - --v=4
+            {{- else if eq .Values.log.level "info" }}
+            - --v=0
```
nckturner: In case the user does want to set the number, should we just pass through in an else block?: 
```
{{- else }}
- --v={{ .Values.log.level }}
{{- end }}
```

kiranmeduri: we can add it later, I am not sure if that is useful anyways because we are not strictly following numbered logging.

PR #75, opened 2020-03-04
upgrade chart to support helm 3

willejs:
This upgrades the chart to support helm3. This in turn means that you don't have to run any manual steps before helm apply to create CRDs.

### DIFF
--- a/stable/aws-calico/Chart.yaml
+++ b/stable/aws-calico/Chart.yaml
@@ -1,7 +1,7 @@
-apiVersion: v1
+apiVersion: v2
 description: A Helm chart for installing Calico on AWS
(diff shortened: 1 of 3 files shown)


stefanprodan:
You do need to run the manual step on upgrade. Helm v3 applies the CRDs only on install. Imagine Calico changes something in the CRD and after an upgrade it will crash since only the deployments will be updated. See https://github.com/helm/helm/issues/7735

I propose we close this PR as it breaks the upgrade. cc @jaypipes

jaypipes:
Agree with @stefanprodan, probably best to close this PR due to manual steps still needing to be taken on upgrade.

PR #81, opened 2020-03-11
new NTH value for dnsPolicy

MrpPolo:
Issue #, if available:
N/A

Description of changes:
Due to daemonset `hostNetwork` setting is `true` , if I want resolve service dns like elasticsearch inside k8s for webhook post , I need to change `dnsPolicy` from `ClusterFirst` to `ClusterFirstWithHostNet`. so new Helm value for this situation.

### DIFF
--- a/stable/aws-node-termination-handler/Chart.yaml
+++ b/stable/aws-node-termination-handler/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
-version: 0.7.0
+version: 0.7.1
 appVersion: 1.3.0
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
--- a/stable/aws-node-termination-handler/values.yaml
+++ b/stable/aws-node-termination-handler/values.yaml
@@ -82,3 +82,5 @@ serviceAccount:
 rbac:
   # rbac.pspEnabled: `true` if PodSecurityPolicy resources should be created
   pspEnabled: true
+
+dnsPolicy: "ClusterFirstWithHostNet"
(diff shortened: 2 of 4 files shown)


stable/aws-node-termination-handler/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
 version: 0.6.0
-appVersion: 1.2.0
```
bwagner5: this version should not be updated, just the regular `version` field above it for helm.  Also, we have a pending PR for v1.3.0 release of node termination handler so this change should go in after that w/ version `0.7.1`

MrpPolo: fixed

bwagner5: nice, should just need to rebase now.

stable/aws-node-termination-handler/values.yaml
```diff
@@ -81,3 +81,5 @@ serviceAccount:
 rbac:
   # rbac.pspEnabled: `true` if PodSecurityPolicy resources should be created
   pspEnabled: true
+
+dnsPolicy: "ClusterFirst"
```
bwagner5: Actually, I think it's best to make the default `ClusterFirstWithHostNet`. Can you update this and the readme default in the table

stable/aws-node-termination-handler/Chart.yaml
MrpPolo: rebased

stable/aws-node-termination-handler/values.yaml
MrpPolo: change `dnsPolicy ` default value to `ClusterFirstWithHostNet`

stable/aws-node-termination-handler/Chart.yaml
bwagner5: looks like the Chart.yaml version got wiped away on rebase. If you update version to `0.7.1`, we should be good to go :)

MrpPolo: done

bwagner5: awesome! thanks!

PR #90, opened 2020-03-16
aws-node-termination-handler: restrict filesystem to readonly and runas non-root 

bwagner5:
## Issue #, if available:
N/A

## Description of changes:
Locks down some access from NTH:

 - Makes the root container filesystem read only
- Runs as non-root
- Disallow privilege escalation explicitly
- change uptime volume mount to readonly

### DIFF
--- a/stable/aws-node-termination-handler/values.yaml
+++ b/stable/aws-node-termination-handler/values.yaml
@@ -4,9 +4,13 @@
 
 image:
   repository: amazon/aws-node-termination-handler
-  tag: v1.3.0
+  tag: v1.3.1
   pullPolicy: IfNotPresent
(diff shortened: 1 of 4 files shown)


stable/aws-node-termination-handler/values.yaml
```diff
@@ -8,6 +8,9 @@ image:
   pullPolicy: IfNotPresent
   pullSecrets: []
 
+runAsUserID: 1000
```
jaypipes: Recommend putting these under a `securityContext` subfield in order to group related values appropriately.

bwagner5:
This patch should be merged first: https://github.com/aws/eks-charts/pull/93

PR #123, opened 2020-05-05
NTH: parameterize os and arch node selector key

bwagner5:
Issue #, if available:
https://github.com/aws/aws-node-termination-handler/pull/149

Description of changes:
 - The OS and Arch node selector keys are moving out of beta. In order to not break users on older kubernetes versions, the node selector keys have been parameterized w/ the default being the beta keys since it still works on K8s 1.11 - 1.18 . 

- Add CORDON_ONLY parameter 

- Add METADATA_TRIES parameter which was added to NTH a while ago but missed putting in the helm chart


haugenj:
Should we add `webhookProxy` into this pr?

bwagner5:
I think I'll hold off until the tests are merged in the main NTH repo

PR #136, opened 2020-05-18
Adding condition for windows nodes

kavichu:
- Use condition to enable windows node selector affinity
- On windows, disable parts of the template that did not work when running, unix's behavior not supported on windows.


bwagner5:
Wow, sorry for the delay! This one must have fallen through the cracks.  I don't think this is needed anymore? If it is, can you rebase and submit this PR on the [aws-node-termination-handler](https://github.com/aws/aws-node-termination-handler/tree/main/config/helm/aws-node-termination-handler) repo? We generally do all development in that repo and then sync it over to eks-charts when we do releases. That way we can run our full set of end-to-end tests.

PR #104, opened 2020-04-08
Adding chart for aws-for-fluent-bit

mforutan:
Signed-off-by: Mohammad Forutan <mforutan@users.noreply.github.com>

Closes #101

Description of changes:

Adding a new chart for [aws-for-fluent-bit](https://github.com/aws/aws-for-fluent-bit)

PR #150, opened 2020-05-30
feat: update aws fluent bit

ArchiFleKs:
* add toleration and node selector
* bump to latest version

Signed-off-by: Kevin Lefevre <lefevre.kevin@gmail.com>

Issue #, if available:

Description of changes:

### DIFF
--- a/stable/aws-for-fluent-bit/Chart.yaml
+++ b/stable/aws-for-fluent-bit/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: aws-for-fluent-bit
 description: A Helm chart to deploy aws-for-fluent-bit project
-version: 0.1.2
-appVersion: 2.2.0
+version: 0.2.0
+appVersion: 2.3.0
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
 sources:
--- a/stable/aws-for-fluent-bit/values.yaml
+++ b/stable/aws-for-fluent-bit/values.yaml
@@ -102,8 +102,12 @@ resources:
   limits:
     memory: 500Mi
   requests:
-    cpu: 500m
-    memory: 500Mi
+    cpu: 50m
+    memory: 50Mi
+
+nodeSelector: {}
+
+tolerations: []
 
 ## Assign a PriorityClassName to pods if set
 # priorityClassName: system-node-critical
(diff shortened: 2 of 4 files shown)


jtackaberry:
Limit of 50MB looks too restrictive.  Here's a deployment of 2.3.0 sending to CloudWatch Logs after a bit of load:

```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
    1 root      20   0  639184  89104  37996 S  29.0  2.2   0:27.15 fluent-bit
```

ArchiFleKs:
I left the limit at 500M I just changed the request of 500M which seem like an excessive lock of resources by default.

jtackaberry:
Ah, indeed, sorry, I'd misread the patch.  Though perhaps a request of 75MB may be more closer to the mark?  Even without any load, my fluent-bit instance comes out of the gate with 65MB.

stable/aws-for-fluent-bit/values.yaml
```diff
@@ -102,8 +102,12 @@ resources:
   limits:
     memory: 500Mi
   requests:
-    cpu: 500m
-    memory: 500Mi
+    cpu: 50m
+    memory: 50Mi
```
mforutan, line 106: Please check [this comment](https://github.com/aws/eks-charts/pull/104#discussion_r406160004) in the initial PR for the reason of the existing limits
cc: @max-rocket-internet

ArchiFleKs: I might be wrong but i'm not sure setting limit=request prevents oomkilled anyway. And since fluent bit is supposed to be resources efficient more than fluentd I think it is a bit excessive to "waste" 500Mi of RAM on a node which cannot be used for other pods in that case the gains compared to fluentd are lost.

But we can leave it this way and let the users change the default values anyway.

stable/aws-for-fluent-bit/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: aws-for-fluent-bit
 description: A Helm chart to deploy aws-for-fluent-bit project
-version: 0.1.2
-appVersion: 2.2.0
+version: 0.2.0
+appVersion: 2.3.0
```
yorinasub17, line 5: Doesn't the image tag need to be updated as well for this change? https://github.com/aws/eks-charts/blob/master/stable/aws-for-fluent-bit/values.yaml#L7

PR #151, opened 2020-05-31
Fix service account name for ClusterRoleBinding in AWS for Fluent Bit

jtackaberry:
In `aws-for-fluent-bit`, the service account name is configurable but `clusterrolebinding.yaml` incorrectly assumes `fullname` rather than `serviceAccountName`.  This pull request fixes this small error.

### DIFF
--- a/stable/aws-for-fluent-bit/Chart.yaml
+++ b/stable/aws-for-fluent-bit/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-for-fluent-bit
 description: A Helm chart to deploy aws-for-fluent-bit project
-version: 0.1.1
+version: 0.1.2
 appVersion: 2.2.0
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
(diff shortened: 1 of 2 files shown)


jtackaberry:
> you need to bump the version though.

@mforutan oh, sure.  I have another pull request, and there are other pull requests pending as well, how do we reconcile conflicting chart versions?  I guess I'm not sure what value to put as it would seem to depend on the order in which you merge the pull requests.

Sorry if that's a newb question. :)

mforutan:
> @mforutan oh, sure. I have another pull request, and there are other pull requests pending as well, how do we reconcile conflicting chart versions? I guess I'm not sure what value to put as it would seem to depend on the order in which you merge the pull requests.
> 
> Sorry if that's a newb question. :)

I would bump the version base on whatever version in the current master, so it would be clear that what is my intention and to tell the reviewers that what kind of change it is, (e.g small change with minor version change or a change that needs major version upgrade), the actual version will be determined at the time of the merge as a conflict which need to be fixed, thanks

jtackaberry:
@mforutan thanks for the explanation, makes sense.  I've updated the version.

jtackaberry:
Sorry about the newline addition.  vim automatically added it and I was a bit hasty in the commit.  Let me know if it's a problem and I'll tweak.

PR #152, opened 2020-05-31
Add PodSecurityPolicy to AWS for Fluent Bit

jtackaberry:
The Helm chart for AWS for Fluent Bit assumes an environment with a permissive default PodSecurityPolicy, which works for out-of-the-box EKS clusters due to the default `eks.privileged` policy, but not when things are tightened up with a restrictive default PSP.  This Helm chart should include a tailored PSP to support AWS for Fluent Bit.

This pull request adds such a PSP and updates the ClusterRole to grant permission to use it.  The assumption of `policy` API group rather than `extensions` does mean that K8s 1.14 is required, however since that's the earliest version supported by EKS as of this writing, the assumption seemed reasonable.

### DIFF
--- a/stable/aws-for-fluent-bit/templates/clusterrole.yaml
+++ b/stable/aws-for-fluent-bit/templates/clusterrole.yaml
@@ -9,4 +9,9 @@ rules:
       - namespaces
       - pods
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["policy"]
+    resources: ["podsecuritypolicies"]
+    verbs: ["use"]
+    resourceNames:
+      - {{ include "aws-for-fluent-bit.fullname" . }}
 {{- end -}}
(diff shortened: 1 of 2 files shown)


mforutan:
Please bump the [chart version](https://github.com/aws/eks-charts/blob/1203e5d3087754df6a0ca93a7b8fa33807d717cf/stable/aws-for-fluent-bit/Chart.yaml#L4)

jtackaberry:
@mforutan yep, ack, so bumped.  Thanks.

jaypipes:
@jtackaberry you will need to rebase here and fix conflicts. I merged your smaller change with bumped the fluent-bit chart version to `0.1.3` (thank you!). This PR bumps it to `0.2.0`. Is that for a particular reason? If not, I'd be cool with you setting chart version to `0.1.3` in this branch and we'll merge both the kinesis partition size change and this addition into a `0.1.3` fluent-bit chart version.

jtackaberry:
> This PR bumps it to `0.2.0`. Is that for a particular reason?

Not a compelling reason. :)  I thought perhaps the addition of a new resource to the chart might warrant a bit more of a version bump.

> If not, I'd be cool with you setting chart version to 0.1.3 in this branch and we'll merge both the kinesis partition size change and this addition into a 0.1.3 fluent-bit chart version.

Sounds great.  I've rebased this PR.  Thanks @jaypipes!

Issue #165, opened 2020-06-12
App Mesh Gateway 


stefanprodan:
Create a chart for the App Mesh Gateway allowing users to configure NLB/ELB, auto scaling, taints and affinity. The `appmesh-gateway` chart should be added to the [eksctl App Mesh profile](https://github.com/weaveworks/eks-appmesh-profile).

PR #168, opened 2020-06-13
Add App Mesh Gateway chart

stefanprodan:
Fix: #165 

Tested on EKS 1.16 with App Mesh preview enabled (App Mesh Manager RC4).

The chart sets the label `app.kubernetes.io/component: "appmesh-gateway"` on all gateway pods, this label should be used to build a dedicated Grafana dashboard for monitoring the gateway instances running on a cluster.

### DIFF
--- a/stable/appmesh-gateway/README.md
+++ b/stable/appmesh-gateway/README.md
@@ -78,6 +78,7 @@ Parameter | Description | Default
 `image.repository` | image repository | `840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-envoy`
 `image.tag` | image tag | `<VERSION>`
 `image.pullPolicy` | image pull policy | `IfNotPresent`
+`skipImageOverride` | when enabled the App Mesh injector will not override the Envoy image | `false`
 `service.type` | service type  | `LoadBalancer`
 `service.port` | service port  | `80`
 `service.annotations` | service annotations | NLB load balancer type
--- a/stable/appmesh-gateway/README.md
+++ b/stable/appmesh-gateway/README.md
@@ -1,11 +1,28 @@
 # App Mesh Gateway
 
-App Mesh Gateway Helm chart for Kubernetes
+App Mesh Gateway Helm chart for Kubernetes. 
 
 ## Prerequisites
 
 * App Mesh CRDs
 * App Mesh Manager >= 1.0.0
+
+**Note** App Mesh Gateway is a release candidate and can be used by
+enabling App Mesh preview features (available only in us-west-2 region).
+
+To enable the preview features:
+
+* When configuring IAM policies, use `appmesh-preview` as the service name instead of `appmesh`
+* Install the App Mesh CRDs with:
+```sh
+kubectl apply -k github.com/aws/eks-charts/stable/appmesh-controller//crds?ref=preview
+```
+* Install the App Mesh Controller chart from the preview branch
+* When configuring pods, add the following annotation so Envoy sidecars point to the preview as well:
+```yaml
+annotations:
+  appmesh.k8s.aws/preview: enabled
+```
 
 ## Installing the Chart
 
--- a/stable/appmesh-gateway/README.md
+++ b/stable/appmesh-gateway/README.md
@@ -1,0 +1,73 @@
+# App Mesh Gateway
+
+App Mesh Gateway Helm chart for Kubernetes
+
+## Prerequisites
+
+* App Mesh CRDs
+* App Mesh Manager >= 1.0.0
+
+## Installing the Chart
+
+Add the EKS repository to Helm:
+
+```sh
+helm repo add eks https://aws.github.io/eks-charts
+```
+
+Create a namespace with injection enabled:
+
+```sh
+kubectl create ns appmesh-ingress
+kubectl label namespace appmesh-ingress appmesh.k8s.aws/sidecarInjectorWebhook=enabled
+```
+
+Deploy the App Mesh Gateway in the `appmesh-ingress` namespace:
+
+```sh
+helm upgrade -i appmesh-gateway eks/appmesh-gateway \
+--namespace appmesh-ingress
+```
+
+Find the NLB address:
+
+```sh
+kubectl get svc appmesh-gateway -n appmesh-ingress
+```
+
+The [configuration](#configuration) section lists the parameters that can be configured during installation.
+
+## Uninstalling the Chart
+
+To uninstall/delete the `appmesh-gateway` deployment:
+
+```console
+$ helm delete appmesh-gateway -n appmesh-ingress
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following tables lists the configurable parameters of the chart and their default values.
+
+Parameter | Description | Default
+--- | --- | ---
+`image.repository` | image repository | `840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-envoy`
+`image.tag` | image tag | `<VERSION>`
+`image.pullPolicy` | image pull policy | `IfNotPresent`
+`service.type` | service type  | `LoadBalancer`
+`service.port` | service port  | `80`
+`service.annotations` | service annotations | NLB load balancer type
+`service.externalTrafficPolicy` | when set to `Local` it preserves the client source IP  | `Cluster`
+`appmesh.gateway` | create a `VirtualGateway` object  | `true`
+`appmesh.preview` | enable App Mesh Preview (us-west-2 only)  | `false`
+`resources.requests/cpu` | pod CPU request | `100m`
+`resources.requests/memory` | pod memory request | `64Mi`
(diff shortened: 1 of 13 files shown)


stable/appmesh-gateway/templates/gateway.yaml
```diff
@@ -0,0 +1,20 @@
...
+{{ include "appmesh-gateway.labels" . | indent 4 }}
+spec:
+  namespaceSelector:
+    matchLabels:
+      appmesh.k8s.aws/sidecarInjectorWebhook: enabled
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: {{ template "appmesh-gateway.fullname" . }}
+  listeners:
+    - portMapping:
```
bcelenza, line 17: Would access logs to /dev/stdout be a nice default here? Also, would be nice to have an option for TLS termination (not a blocker, assume most folks will terminate @ NLB and not re-encrypt to gateway).

stefanprodan: @bcelenza I've added the access logs

stable/appmesh-gateway/templates/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+            - podAffinityTerm:
+                labelSelector:
+                  matchLabels:
+                    app.kubernetes.io/name: {{ include "appmesh-gateway.name" . }}
+                topologyKey: kubernetes.io/hostname
+              weight: 100
+      {{- end }}
+      containers:
+      - name: envoy
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
bcelenza, line 44: Default `image` should be `injector-envoy-image` to match the [associated mutator changes](https://github.com/aws/aws-app-mesh-controller-for-k8s/commit/715b59a98280858313e5707f601b5326ca3b58cc), so the Envoy from the appmesh-manager is automatically injected.

stable/appmesh-gateway/templates/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        ports:
+        - containerPort: 8088
+          name: http
+          protocol: TCP
+        - containerPort: 9901
+          name: http-prom
+          protocol: TCP
```
bcelenza, line 52: `prom` as in prometheus? 9901 is the Envoy admin port for the App Mesh managed images. Probably worth clarifying that here in the name.

stable/appmesh-gateway/templates/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+        livenessProbe:
+          tcpSocket:
+            port: http-prom
+        readinessProbe:
+          initialDelaySeconds: 5
+          tcpSocket:
+            port: http-prom
```
bcelenza, line 59: Both of these will be considered "good" pretty much as soon as the Envoy starts, given it's the admin port. We'd probably want to clarify this with an actual check for bootstrapped configuration, e.g. [this command](https://github.com/aws/aws-app-mesh-examples/blob/master/walkthroughs/tls-with-acm/infrastructure/ecs-service.yaml#L116-L118).

stable/appmesh-gateway/templates/psp.yaml
```diff
@@ -0,0 +1,57 @@
...
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
+spec:
+  privileged: false
+  hostIPC: false
+  hostNetwork: false
+  hostPID: false
+  readOnlyRootFilesystem: false
+  allowPrivilegeEscalation: false
+  allowedCapabilities:
+    - '*'
```
bcelenza, line 18: Just to confirm, this will enable the pod with `CAP_NET_BIND_SERVICE`? This is a [known edge](https://docs.aws.amazon.com/app-mesh/latest/userguide/troubleshooting-setup.html#virtual-gateway-low-ports) that's unique about gateway, given we bind to the specified port instead of 15000 for ingress.

stable/appmesh-gateway/templates/deployment.yaml
fawadkhaliq: Correction to what Brian said: it is injected by default but you can skip it (to use user provided envoy image) using the annotation `appmesh.k8s.aws/virtualGatewaySkipImageOverride`. Here are the details [1]

[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/virtualgateway_envoy.go#L140

stable/appmesh-gateway/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+```sh
+helm repo add eks https://aws.github.io/eks-charts
+```
+
+Create a namespace with injection enabled:
+
+```sh
+kubectl create ns appmesh-ingress
+kubectl label namespace appmesh-ingress appmesh.k8s.aws/sidecarInjectorWebhook=enabled
```
fawadkhaliq, line 22: Would be good to make `mesh` configurable since a cluster may have multiple meshes and each mesh having it's own set of virtual gateways

stable/appmesh-gateway/templates/gateway.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: VirtualGateway
+metadata:
+  name: {{ template "appmesh-gateway.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "appmesh-gateway.labels" . | indent 4 }}
+spec:
+  namespaceSelector:
+    matchLabels:
+      appmesh.k8s.aws/sidecarInjectorWebhook: enabled
```
fawadkhaliq, line 12: Is there a reason for using pod injection label as virtual gateway namespaceSelector? It'd be better to make to use an explicit label instead to avoid selecting all the namespaces across all the meshes if customers use `appmesh.k8s.aws/sidecarInjectorWebhook: enabled`

stable/appmesh-gateway/templates/gateway.yaml
```diff
@@ -0,0 +1,24 @@
...
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: {{ template "appmesh-gateway.fullname" . }}
+  listeners:
+    - portMapping:
+        port: 8088
```
fawadkhaliq, line 18: Should we make the port configurable?

stable/appmesh-gateway/templates/deployment.yaml
stefanprodan: I've set the `prom` suffix to signal that this port is used by the App Mesh Prometheus, I'll change it.

stable/appmesh-gateway/templates/deployment.yaml
stefanprodan: Wouldn't that cause all pods to crash loop when the App Mesh xDS server has an outage? My assumption is that Envoy should keep working with the last know configuration, do we want to have a cascading failure?

stable/appmesh-gateway/templates/psp.yaml
stefanprodan: Yes that should be the case.

stable/appmesh-gateway/templates/gateway.yaml
stefanprodan: I think that's an advanced use-case, the chart tries to make it easy for those that are running a single ingress. I would expect users that what to run multiple gateways for multiple meshes to set `appmesh.gateway=false` and create their own objects.

stable/appmesh-gateway/templates/gateway.yaml
stefanprodan: Why would a user change it? The ports that matter in my opinion are the ones exposed via NLB.

stable/appmesh-gateway/templates/deployment.yaml
stefanprodan: I've added `skipImageOverride` to chart options.

stable/appmesh-gateway/templates/gateway.yaml
fawadkhaliq: Sounds good

stable/appmesh-gateway/templates/gateway.yaml
fawadkhaliq: The idea was for users who would like to play with different knobs exposed. Not needed as such

stable/appmesh-gateway/templates/deployment.yaml
bcelenza: So long as the Envoy has bootstrapped the initial configuration, it will respond back with the `LIVE` state. This is true even if the xDS server has crashed and Envoy is retrying/backing off on the connection. 

It would be the case that an Envoy which cannot bootstrap due to an xDS outage would be considered unhealthy, but that would be expected.

stefanprodan: Ok I've added the live check to the liveness probe, thanks for the explanation.

PR #173, opened 2020-06-16
Maintain one chart for App Mesh controller

fawadkhaliq:
*Description:*
We have been using manager and controller interchangeably for the App Mesh controller. This causes confusion in the documents. Instead of keeping two separate helm charts, we will 
 maintain only one `appmesh-controller` chart moving forward and it will carry the new releases. 

*Changes include:*
- Rename manager to controller in appmesh-manager chart
- Replace appmesh-controller with appmesh-manager (now appmesh-controller)

*Testing:*
- Tested installation and end to end controller with both rc5 and rc-preview tags


stable/appmesh-controller/Chart.yaml
```diff
@@ -1,19 +1,16 @@
 apiVersion: v1
 name: appmesh-controller
 description: App Mesh controller Helm chart for Kubernetes
-version: 0.6.1
-appVersion: 0.5.0
+version: 0.6.2
```
M00nF1sh, line 4: let's bump this chart version to a new major version like v1.0.0 as well

stable/appmesh-controller/README.md
```diff
@@ -4,6 +4,20 @@ App Mesh controller Helm chart for Kubernetes
...
 
+**Note** App Mesh controller is a release candidate. Please use it for testing purpose only as it has backward incompatible changes with v0.5.0. Upgrade instructions will be provided before the final release.
+
+**Note** If you wish to use App Mesh preview features, you can add `-preview` to the published image. e.g. `v1.0.0-rc4-preview`
+Two important things when using App Mesh preview:
+1. When configuring IAM policies, use `appmesh-preview` as the service name instead of `appmesh`. See the example JSON below.
+2. When configuring pods, add the following annotation so Envoy sidecars point to the preview as well:
+```
+annotations:
+  appmesh.k8s.aws/preview: "true"
```
M00nF1sh, line 15: i believe we changed from true to enabled

fawadkhaliq: Yes, thank you for catching this! It was marked to update and missed

stable/appmesh-controller/ci/values.yaml
```diff
@@ -1,7 +1,7 @@
-# CI testing values for appmesh-controller.
+# CI testing values for appmesh-controller
 
 region: us-west-2
 image:
-  repository: stefanprodan/aws-app-mesh-controller-for-k8s
-  tag: v0.2.0
+  repository: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller
+  tag: v1.0.0-rc5
```
bcelenza, line 6: Shouldn't the default be the current stable version (0.2.0) with an option to opt-in to the release candidate?

bcelenza: Also, image repository should be regionalized.

fawadkhaliq: The charts will be not published by this check-in so current users will continue to get v0.2.0. Charts will be published when we build a new tag at the release time. Keeping v0.2.0 here will not help as all the configuration changes/crds are not compatible with v0.2.0 image

fawadkhaliq: > Also, image repository should be regionalized.

That's a good idea to make it configurable. I will mark this for a follow up change

PR #188, opened 2020-06-24
aws-calico: add podSecurityPolicy objects

KYannick:
add parameters to enable the creation of podsecuritypolicy objects for calico components

Signed-off-by: Yannick Kint <yannickkint@gmail.com>

Issue #, if available:

Description of changes:
The Helm chart for AWS for calico assumes an environment with a permissive default PodSecurityPolicy, which works for out-of-the-box EKS clusters due to the default eks.privileged policy, but not when things are tightened up with a restrictive default PSP. This Helm chart should include a tailored podSecurityPolicy to allow calico to work in all environments.

creation of the objects can be enabled via the values of the helm chart, by default they will not be created.

### DIFF
--- a/stable/aws-calico/Chart.yaml
+++ b/stable/aws-calico/Chart.yaml
@@ -2,6 +2,6 @@ apiVersion: v1
 description: A Helm chart for installing Calico on AWS
 website: https://docs.aws.amazon.com/eks/latest/userguide/calico.html
 name: aws-calico
-version: 0.2.2
+version: 0.3.0
 appVersion: 3.8.1
 icon: https://www.projectcalico.org/wp-content/uploads/2019/09/Calico_Logo_Large_Calico.png
--- a/stable/aws-calico/README.md
+++ b/stable/aws-calico/README.md
@@ -51,6 +51,7 @@ The following table lists the configurable parameters for this chart and their d
 | `fullnameOverride`      | Override the fullname of the chart                      | `calico`                        |
 | `serviceAccount.name`   | The name of the ServiceAccount to use                   | `nil`                           |
 | `serviceAccount.create` | Specifies whether a ServiceAccount should be created    | `true`                          |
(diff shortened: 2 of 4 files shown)


KYannick:
@fawadkhaliq could you take a look at this PR? Or if you are not the correct person, could you add the correct reviewers for this PR? I know there is a conflict on the version in the chart.yaml, but I rather first get feedback the rest of the changes is ok, before solving it.

stable/aws-calico/Chart.yaml
```diff
@@ -2,6 +2,6 @@ apiVersion: v1
 description: A Helm chart for installing Calico on AWS
 website: https://docs.aws.amazon.com/eks/latest/userguide/calico.html
 name: aws-calico
-version: 0.2.2
+version: 0.3.0
```
jaypipes, line 5: set this to 0.3.1 since I already merged https://github.com/aws/eks-charts/pull/228. We'll cut a release of the 0.3.1 calico chart with this included as well as the fix from 228

KYannick: I merged the upstream back in my branch, so it is at 0.3.1 now.

stable/aws-calico/README.md
```diff
@@ -52,6 +52,7 @@ The following table lists the configurable parameters for this chart and their d
 | `fullnameOverride`      | Override the fullname of the chart                      | `calico`                        |
 | `serviceAccount.name`   | The name of the ServiceAccount to use                   | `nil`                           |
 | `serviceAccount.create` | Specifies whether a ServiceAccount should be created    | `true`                          |
+| `podSecurityPolicy.create` | Specifies whether podSecurityPolicy and related rbac objects should be created    | `false`                          |
```
jaypipes, line 55: Note that @casparrolfe cleaned up the formatting of this markdown table in #233. So you might want to do the same in a future commit. Merging this now.

Issue #57, opened 2019-12-20
Handle helm upgrade of AWS VPC CNI more gracefully


realvz:
According the documentation in aws-VPC-CNI helm chart,
“ If you receive an error similar to Error: release aws-vpc-cni failed: <resource> "aws-node" already exists, simply rerun the above command.”

This causes issues with automation. This should be handled automatically in the helm chart and not require a rerun of the command by user.

max-rocket-internet:
> This should be handled automatically in the helm chart

For sure but how? To my knowledge, it's not possible blindly overwrite existing resources without errors. But the behaviour has also changed in the last couple releases of Helm.

3oris:
The a above quoted statement:

> If you receive an error similar to Error: release aws-vpc-cni failed: "aws-node" already exists, simply rerun the above command.

is just not true:

jaypipes:
Thanks for the heads up @3oris, this is embarrassing indeed. I will work on getting an upgrade job placed into the VPC CNI system once we move the chart sources in there.

mohsen0:
I removed the daemon set object before applying the change and managing CNI with helm.
after that everything went smoothly

3oris:
@mohsen0 for me it doesn't. Helm complains about the pre-existing CRD next. I surely know that I could delete all related resources and then apply the chart. For me the question more or less is: Should an EKS cluster come with a pre-installed CNI plugin if after all the CNI plugin wont be part of the "managed service" promise. Now the thing comes with an unmanaged component out of the box that leaves me manage it myself, but adequate means of automation (in terms of CD) cannot be applied, and I have to work around that.

stefanprodan:
The CNI chart is not compatible with Helm v3, we should move the CRD to the `crds` dir, here is an example: https://github.com/aws/eks-charts/tree/master/stable/appmesh-controller

mohsen0:
@3oris I understand there is room for more improvement and clarify, But I suggest a more constructive, positive approach. then your contributions will be much more appreciated.

max-rocket-internet:
CRD issue aside, you can just install the chart, it fails, then purge it, then reinstall and all is done. Obviously deletes all the resources and then recreates them with the chart. I've done this method many times now and it works fine without any interruption.

max-rocket-internet:
> According the documentation in aws-VPC-CNI helm chart,
> “ If you receive an error similar to Error: release aws-vpc-cni failed: "aws-node" already exists, simply rerun the above command.”

[...]

PR #192, opened 2020-07-02
allow original match label on daemonset to enable helm adoption

willejs:
Description of changes:
This PR enables the adoption of the existing aws-node resources by allowing you to retain the original match labels, and results in a successful adoption.

The match label on a daemonset is an immutable field. When trying to adopt the existing aws-node resources for aws-vpc-cni it will diff on these matchLabels and attempt to change them, resulting in an error.

This is somewhat related to issue https://github.com/aws/eks-charts/issues/57.

### DIFF
--- a/stable/aws-vpc-cni/README.md
+++ b/stable/aws-vpc-cni/README.md
@@ -42,6 +42,7 @@ The following table lists the configurable parameters for this chart and their d
 | `image.pullPolicy`      | Container pull policy                                   | `IfNotPresent`                      |
 | `image.override`        | A custom docker image to use                            | `nil`                               |
 | `imagePullSecrets`      | Docker registry pull secret                             | `[]`                                |
+| `originalMatchLabels`   | Use the original daemonset matchLabels                  | `false`                             |
 | `nameOverride`          | Override the name of the chart                          | `aws-node`                          |
 | `nodeSelector`          | Node labels for pod assignment                          | `{}`                                |
 | `podSecurityContext`    | Pod Security Context                                    | `{}`                                |
@@ -61,3 +62,7 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
 ```shell
 $ helm install --name aws-vpc-cni --namespace kube-system eks/aws-vpc-cni --values values.yaml
 ```
+
+## Adopting the existing aws-node resources in an EKS cluster
+
+If you do not want to delete the existing aws-node resources in your cluster that run the aws-vpc-cni and then install this helm chart, you can adopt the resources into a release instead. This process is highlighted in this [PR comment](https://github.com/aws/eks-charts/issues/57#issuecomment-628403245). Once you have annotated and labeled all the resources this chart specifies, enable the `originalMatchLabels` flag on the helm release and run an update. If you have been careful this should not diff and leave all the resources unmodified and now under management of helm.
--- a/stable/aws-vpc-cni/README.md
+++ b/stable/aws-vpc-cni/README.md
@@ -62,6 +62,7 @@ $ helm install --name aws-vpc-cni --namespace kube-system eks/aws-vpc-cni --valu
 If you do not want to delete the existing aws-node resources in your cluster that run the aws-vpc-cni and then install this helm chart, you can adopt the resources into a release instead. This process is highlighted in this [PR comment](https://github.com/aws/eks-charts/issues/57#issuecomment-628403245). Once you have annotated and labeled all the resources this chart specifies, enable the `originalMatchLabels` flag on the helm release and run an update. If you have been careful this should not diff and leave all the resources unmodified and now under management of helm.
 
 Here is an example script to modify the existing resources:
+
 WARNING: Substitute YOUR_HELM_RELEASE_NAME_HERE with the name of your helm release.
 ```
 #!/usr/bin/env bash
--- a/stable/aws-vpc-cni/README.md
+++ b/stable/aws-vpc-cni/README.md
@@ -70,7 +70,7 @@ set -euo pipefail
 
 # don't import the crd. Helm cant manage the lifecycle of it anyway.
 for kind in daemonSet clusterRole clusterRoleBinding serviceAccount; do
-  echo "setting annotations and labels on $kind"
+  echo "setting annotations and labels on $kind/aws-node"
   kubectl -n kube-system annotate --overwrite $kind aws-node meta.helm.sh/release-name=YOUR_HELM_RELEASE_NAME_HERE
   kubectl -n kube-system annotate --overwrite $kind aws-node meta.helm.sh/release-namespace=kube-system
   kubectl -n kube-system label --overwrite $kind aws-node app.kubernetes.io/managed-by=Helm
--- a/stable/aws-vpc-cni/README.md
+++ b/stable/aws-vpc-cni/README.md
@@ -20,13 +20,7 @@ To install the chart with the release name `aws-vpc-cni` and default configurati
 $ helm install --name aws-vpc-cni --namespace kube-system eks/aws-vpc-cni
 ```
 
-To install into an EKS cluster where the CNI is already installed, you can run:
-
-```shell
-helm upgrade --install --recreate-pods --force aws-vpc-cni --namespace kube-system eks/aws-vpc-cni
-```
-
-If you receive an error similar to `Error: release aws-vpc-cni failed: <resource> "aws-node" already exists`, simply rerun the above command.
+To install into an EKS cluster where the CNI is already installed, see [this section below](##-Adopting-the-existing-aws-node-resources-in-an-EKS-cluster)
 
 ## Configuration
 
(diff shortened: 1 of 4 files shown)


stable/aws-vpc-cni/templates/daemonset.yaml
```diff
@@ -9,8 +9,12 @@ spec:
 {{ toYaml .Values.updateStrategy | indent 4 }}
   selector:
     matchLabels:
+{{- if .Values.originalMatchLabels }}
+      k8s-app: aws-node
+{{- else }}
```
mohsen0, line 14: we could keep the extra labels and we don't have to exclude them.
it will help to have consistent labelling across workloads

willejs: Hi @mohsen0 👋 
I am not sure I understand what you mean here?
The matchLabels field cannot be altered at all without recreating the ds, so it has to be just this single label.

mohsen0: hey

So changing the labels on daemon set object will not cause a recreation, I just checked that by adding a label to a daemon set I have.

What I mean is having it like
```
{{- if .Values.originalMatchLabels }}
      k8s-app: aws-node
{{- end }}
      app.kubernetes.io/name: {{ include "aws-vpc-cni.name" . }}
      app.kubernetes.io/instance: {{ .Release.Name }}
```

so we can have a consistent naming in order to for example to look up logs in kibana or looking up resources.

stable/aws-vpc-cni/README.md
```diff
@@ -61,3 +62,7 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
 ```shell
 $ helm install --name aws-vpc-cni --namespace kube-system eks/aws-vpc-cni --values values.yaml
 ```
+
+## Adopting the existing aws-node resources in an EKS cluster
+
+If you do not want to delete the existing aws-node resources in your cluster that run the aws-vpc-cni and then install this helm chart, you can adopt the resources into a release instead. This process is highlighted in this [PR comment](https://github.com/aws/eks-charts/issues/57#issuecomment-628403245). Once you have annotated and labeled all the resources this chart specifies, enable the `originalMatchLabels` flag on the helm release and run an update. If you have been careful this should not diff and leave all the resources unmodified and now under management of helm.
```
TBBle, line 68: Since we know the resources we are trying to adopt, could we perhaps just give the full set of `kubectl annotate`/`kubectl label` commands?

TBBle:
Since this is fixing the issue that "just rerunning Helm with `--recreate` until it blats the existing resources" doesn't work in current Helm versions, should that instruction (lines 23-30 of README.md) also be removed?

stable/aws-vpc-cni/README.md
willejs: I have added an example script to add these to all the resources, hopefully it will give people some inspiration.

willejs:
@TBBle thanks for the review! I think i have addressed your comments, are they acceptable?

stable/aws-vpc-cni/templates/daemonset.yaml
willejs: No worries! I thought you meant that for a minute, but then I thought you meant to modify the pod labels too, but that wouldn't work either! Hope you are well mate!

stable/aws-vpc-cni/README.md
```diff
@@ -20,13 +20,7 @@ To install the chart with the release name `aws-vpc-cni` and default configurati
...
 ```
 
-To install into an EKS cluster where the CNI is already installed, you can run:
-
-```shell
-helm upgrade --install --recreate-pods --force aws-vpc-cni --namespace kube-system eks/aws-vpc-cni
-```
-
-If you receive an error similar to `Error: release aws-vpc-cni failed: <resource> "aws-node" already exists`, simply rerun the above command.
+To install into an EKS cluster where the CNI is already installed, see [this section below](##-Adopting-the-existing-aws-node-resources-in-an-EKS-cluster)
```
TBBle, line 23: The Anchor link is not working for me in GitHub's Markdown viewer. Unless I'm missing something, the `#-` shouldn't be there, it should be `#Adopting-the-existing-aws-node-resources-in-an-EKS-cluster`

stable/aws-vpc-cni/README.md
```diff
@@ -66,3 +60,18 @@ $ helm install --name aws-vpc-cni --namespace kube-system eks/aws-vpc-cni --valu
...
 If you do not want to delete the existing aws-node resources in your cluster that run the aws-vpc-cni and then install this helm chart, you can adopt the resources into a release instead. This process is highlighted in this [PR comment](https://github.com/aws/eks-charts/issues/57#issuecomment-628403245). Once you have annotated and labeled all the resources this chart specifies, enable the `originalMatchLabels` flag on the helm release and run an update. If you have been careful this should not diff and leave all the resources unmodified and now under management of helm.
+
+Here is an example script to modify the existing resources:
+WARNING: Substitute YOUR_HELM_RELEASE_NAME_HERE with the name of your helm release.
+```
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+for kind in ds clusterRole clusterRoleBinding serviceAccount; do
```
TBBle, line 71: I think we need to annotate the `eniconfigs.crd.k8s.amazonaws.com` CRD too.

Although it's not a great idea, someone using `crd.create=true` (which is the default) will need it annotated to adopt it, as seen at https://github.com/aws/eks-charts/issues/57#issuecomment-587581045.

----

Alternatively, moving the CRD to [Helm v3's approach](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds) (`crds/` directory instead of `templates/`) would remove this need, as Helm would no longer _own_ the CRD, but merely install it if not already present. That requires Helm v2 users to manually install the CRD, which would need to be documented at least [until Helm v2 terminates support in November 2020](https://helm.sh/blog/covid-19-extending-helm-v2-bug-fixes/).

We can see the latter flow in the [`appmesh-controller` chart](https://github.com/aws/eks-charts/tree/master/stable/appmesh-controller) as mentioned at https://github.com/aws/eks-charts/issues/57#issuecomment-587588998

TBBle:
Nice work, thank you for addressing my comments. A cosmetic and non-cosmetic issue arose from the changes, but once those are resolved (or punted into a different Issue/PR as appropriate) it looks good to me.

stable/aws-vpc-cni/README.md
```diff
@@ -61,3 +56,22 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
...
+  echo "setting annotations and labels on $kind"
```
TBBle, line 72: Nit-pick: `$kind/aws-node` so that just the text output is sufficient to know what things were changed. Also matters because the CRD is not named aws-node, so it would be slightly confusing.

stable/aws-vpc-cni/README.md
TBBle: Also, a nit: `daemonSet` instead of `ds`? The rest are the long-form, so this would be nice and consistent.

I'd also capitalise the first letter, as seen in the files themselves, even though they're case-insensitive. I may be in the minority on this though. And it's only documentation.

willejs: I have changed the ds to daemonSet and added a comment.
I don't think this PR should move the existing resources, and I don't think that we should import them as helm cant manage the lifecycle of them anyway.

TBBle: I agree, but right now the chart does try to treat the CRD as a regular resource by default, and will clash with the existing CRD. We may have to tell users adopting resources to set `crd.create=false` so that it leaves the CRD alone.

(Unless Helm v3 has some magic to ignore CRDs in a template. I believed it only ignored them when installed by the crd-install hook used by Helm v2, but otherwise would treat them as normal resources.)

stable/aws-vpc-cni/README.md
```diff
@@ -61,3 +56,23 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
 ```shell
 $ helm install --name aws-vpc-cni --namespace kube-system eks/aws-vpc-cni --values values.yaml
 ```
+
+## Adopting the existing aws-node resources in an EKS cluster
+
+If you do not want to delete the existing aws-node resources in your cluster that run the aws-vpc-cni and then install this helm chart, you can adopt the resources into a release instead. This process is highlighted in this [PR comment](https://github.com/aws/eks-charts/issues/57#issuecomment-628403245). Once you have annotated and labeled all the resources this chart specifies, enable the `originalMatchLabels` flag on the helm release and run an update. If you have been careful this should not diff and leave all the resources unmodified and now under management of helm.
+
+Here is an example script to modify the existing resources:
+WARNING: Substitute YOUR_HELM_RELEASE_NAME_HERE with the name of your helm release.
```
TBBle, line 65: This needs a preceding newline, otherwise it runs into the line above, which doesn't appear to be intended. Also consider marking YOUR_HELM_RELEASE_NAME_HERE with backticks so it gets formatted to match the following code.

willejs: sorted! cheers.

stable/aws-vpc-cni/README.md
willejs: I've added a comment. I think if I move this it will affect existing releases and would be a breaking change which isn't the aim of the pr. Its best left to another pr i feel.

willejs:
@fawadkhaliq could you review and potentially merge this, please?

fawadkhaliq:
Adding the CNI guru @mogren to review this :)

PR #214, opened 2020-07-19
Allow kinesis.partitionKey to be optional

jtackaberry:
The AWS for Fluent Bit chart is presently undeployable when using Kinesis Data Streams as an output.

The root cause is the mandatory nature of `kinesis.partitionKey` combined with https://github.com/aws/amazon-kinesis-streams-for-fluent-bit/issues/30.   `kinesis.partitionKey` needs to be something like `kubernetes.pod_id` or `kubernetes.docker_id` however this isn't supported by the Kinesis streams plugin.

Since there are no other suitable top-level keys to be used as a partition key, and setting it as the empty string prevents Fluent Bit from starting at all, the only other option is to leave `partition_key` undefined in the Fluent Bit OUTPUT spec, but the Helm chart doesn't allow this and defaults to `container_id`.  This results in a steady spew of `Invalid partition key. Failed to find partition_key container_id in log record` errors.

This merge request allows partitionKey to be explicitly unset in user-defined chart values, which results in it not being specified at all in the config map.  This isn't as ideal as fixing the [original problem](https://github.com/aws/amazon-kinesis-streams-for-fluent-bit/issues/30) but at least it allows this Helm chart to be used at all with Kinesis streams.


Fran-ELS-AMS:
👍 
I'm facing the same issue right now.
I hope this PR will get reviewed soon

jtackaberry:
@Fran-ELS-AMS thanks for chiming in -- I hope so too.

We have forked this chart internally and we've made some other improvements as well but I decided not to submit PRs just so as not to create further distractions from blockers like this one (or [this one](https://github.com/aws/eks-charts/pull/152)).

PR #217, opened 2020-07-23
Point to managed policies for required app mesh controller permissions

efe-selcuk:
CreateMesh will use the calling identity to create the SLR in the customer account. Without this, the CreateMesh calls from the controller wouldn't be able to create the SLR, which would manifest as Cloud Map discovery issues. AWSAppMeshFullAccess does contain this permission, so the IRSA instructions work fine.

I'm open to alternatives. For example, relying directly on AWSAppMeshFullAccess rather than listing all the permissions here.

### DIFF
--- a/stable/appmesh-controller/Chart.yaml
+++ b/stable/appmesh-controller/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: appmesh-controller
 description: App Mesh controller Helm chart for Kubernetes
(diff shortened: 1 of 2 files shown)


efe-selcuk:
@fawadkhaliq I can't request reviewer directly but maybe you can take a look.

efe-selcuk:
I should have looked more closely at the CI failure, my mistake. On it.

Issue #125, opened 2020-05-06
Missing parameter for use of STS Regional endpoints


rafaelpereyra:
Hello, 

Using VPC endpoints creates DNS records for the services that include regional names. In order to use the STS regional endpoint it's required to pass en environment variable to the Go SDK (AWS_STS_REGIONAL_ENDPOINTS).

Helm chart should include this option as an override for the default variable value.

fawadkhaliq:
This was fixed here: https://github.com/aws/eks-charts/pull/236

PR #236, opened 2020-08-25
STS regional Endpoint

rafaelpereyra:
Issue #125 

Description of changes:

Added an optional value to specify the aws-account-id param to the controller.  This will fix the STS issues indicated here [328](https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/328), [329](https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/329) and [330](https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/330)

### DIFF
--- a/stable/appmesh-controller/README.md
+++ b/stable/appmesh-controller/README.md
@@ -267,4 +267,5 @@ Parameter | Description | Default
 `enableCertManager` |  Enable Cert-Manager | `false`
 `xray.image.repository` | X-Ray image repository | `amazon/aws-xray-daemon`
 `xray.image.tag` | X-Ray image tag | `latest`
+`accountId` | AWS Account ID must be provided for a VPC without internet access | None
 `env` |  environment variables to be injected into the appmesh-controller pod | `{}`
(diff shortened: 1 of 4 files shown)


stable/appmesh-controller/values.yaml
```diff
@@ -4,6 +4,7 @@
 
 replicaCount: 1
 region: ""
+accountId: ""
```
fawadkhaliq, line 7: Looks good. Can you please also update the README [here](https://github.com/aws/eks-charts/blob/master/stable/appmesh-controller/README.md#configuration)

rafaelpereyra: I added the parameter to the README file as requested. Thanks for the feedback!

stable/appmesh-controller/README.md
```diff
@@ -267,4 +267,5 @@ Parameter | Description | Default
 `enableCertManager` |  Enable Cert-Manager | `false`
 `xray.image.repository` | X-Ray image repository | `amazon/aws-xray-daemon`
 `xray.image.tag` | X-Ray image tag | `latest`
+`accountId` | AWS Account ID must be provided for a VPC without internet access | None
```
fawadkhaliq, line 270: The current description may not be accurate in all cases. Can we please change it to something like: `AWS Account ID for the Kubernetes cluster`

rafaelpereyra: Updated the description as suggested, however I'll recommend to specify parameter is needed for Private VPCs (maybe in the release notes?)

fawadkhaliq: Thanks! +1

PR #242, opened 2020-09-07
CustomResourceDefinition is not beta anymore since 1.16

Kent1:
Issue #, if available:

Description of changes:

### DIFF
--- a/stable/aws-vpc-cni/templates/customresourcedefinition.yaml
+++ b/stable/aws-vpc-cni/templates/customresourcedefinition.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.crd.create -}}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: eniconfigs.crd.k8s.amazonaws.com


mogren:
@Kent1 that's true, but EKS is still supporting 1.14 and 1.15 clusters, so we can't quite remove this yet.

Kent1:
I just saw that aws-vpc-cni 1.6.3 was indeed supported for EKS 1.14, i thought every EKS version had a different aws-vpc-cni version, sorry.
How do you handle versioning here ? I see there no branch or tag ?

mogren:
@Kent1 Currently we do our best to remain backwards compatible when possible, since this CNI is used on AWS outside of EKS as well. For example, we have both [beta and non-beta selectors in the sample configs](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.7/aws-k8s-cni.yaml#L89-L117), since a [1.15 kube-apiserver will still accept 1.13 kubelet](https://kubernetes.io/docs/setup/release/version-skew-policy/#kubelet), and `kubernetes.io/os` and `kubernetes.io/arch` only became GA in 1.14.

Kent1:
Ok thanks for the precisions, i close this PR.

PR #245, opened 2020-09-10
[calico-node] Adds support for pod annotations.

vladlosev:
Description of changes:

This PR adds support for specifying pod annotations on components of `calico-node`. Annotations have many uses. For example many systems such as prometheus or Datadog agents use them to note how metrics should be collected off the pods.


jaypipes (approved):
We should be updating the chart version in chart.yaml, but since I'm merging another PR that does this, I'm good to merge this. Thank you @vladlosev! :)

PR #253, opened 2020-09-17
AWS VPC CNI Chart updates for 1.7.3 release

shaunofneuron:
Issue #, if available:

Description of changes: Changes reflect the new configurations provided by the latest release of aws-vpc-cni 1.7.2. In addition the readiness/liveness probe configurations are exposed to values.yaml.

@mogren There are some big changes here but this should be a good start to getting the helm chart up to date. I've diffed as much as possible against what is defined [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.7/aws-k8s-cni.yaml) and what I have running, you may have a better way of testing changes (please share).


mogren (changes requested):
Thanks a lot @shaunofneuron, this looks great!

I do think you need to bump [`stable/aws-vpc-cni/Chart.yaml`](https://github.com/aws/eks-charts/blob/master/stable/aws-vpc-cni/Chart.yaml) as well though, both `version` and `appVersion`.

Also, note that we did release [v1.7.3](https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.7.3) version yesterday with just [one more commit, affecting the init-container](https://github.com/aws/amazon-vpc-cni-k8s/pull/1212/files).

shaunofneuron:
I'll bump the version and also run through some additional testing, will let you know.

stable/aws-vpc-cni/values.yaml
```diff
@@ -6,19 +6,43 @@
...
+  AWS_VPC_K8S_CNI_LOG_FILE: "/host/var/log/aws-routed-eni/ipamd.log"
   AWS_VPC_K8S_CNI_LOGLEVEL: DEBUG
+  AWS_VPC_K8S_CNI_RANDOMIZESNAT: "prng"
   AWS_VPC_K8S_CNI_VETHPREFIX: eni
-  AWS_VPC_ENI_MTU: "9001"
+  AWS_VPC_K8S_PLUGIN_LOG_FILE: "/var/log/aws-routed-eni/plugin.log"
+  AWS_VPC_K8S_PLUGIN_LOG_LEVEL: DEBUG
+  DISABLE_INTROSPECTION: "false"
+  DISABLE_METRICS: "false"
+  DISABLE_TCP_EARLY_DEMUX: "false"
```
mogren, line 43: Thanks a lot @shaunofneuron, this is awesome. Just one thing here, the `DISABLE_TCP_EARLY_DEMUX` is actually just used by the init container. It's only needed under some special circumstances when using the per pod security group feature.

https://github.com/aws/amazon-vpc-cni-k8s/blob/638c3ca21bde5b5f04a1c5aff228c472d81780f3/config/v1.7/aws-k8s-cni.yaml#L195-L199

tsunamishaun: Totally missed that, good catch

tsunamishaun: Exposed init env and shifted that var into it.

shaunofneuron: Exposed init env and moved var into that.

mogren (approved):
Thanks a lot @shaunofneuron, everything looks good to me. 🚀

shaunofneuron:
You're welcome @mogren ! Thank you for all the good feedback.

stable/aws-vpc-cni/templates/clusterrole.yaml
```diff
@@ -8,17 +8,18 @@ rules:
...
       - namespaces
     verbs: ["list", "watch", "get"]
+  - apiGroups: [""]
+    resources:
+      - nodes
+    verbs: ["list", "watch", "get", "update"]
   - apiGroups: ["extensions"]
     resources:
-      - daemonsets
+      - '*'
```
kishorj, line 24: are there any specific resources?

stable/aws-vpc-cni/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-vpc-cni
-version: 1.0.9
-appVersion: "v1.6.3"
+version: 1.0.10
```
kishorj, line 3: Since there are significant changes, lets make it 1.1.0 instead.

kishorj (commented):
I've some minor comments, changes look good otherwise.

stable/aws-vpc-cni/templates/clusterrole.yaml
mogren: I believe this change was made to match the config in the CNI repo, [v1.7/aws-k8s-cni.yaml](https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.7.3/config/v1.7/aws-k8s-cni.yaml#L46-L52). The change was added in https://github.com/aws/amazon-vpc-cni-k8s/pull/1082 ping @jayanthvn

kishorj: If we know for sure which resources are needed, we can update the helm chart independently of the cni repo. By specifying * we might be allowing more than necessary access.

shaunofneuron: It has been a while since I've had to write policies and not entirely sure where to get the entire list outside of kubectl cache but here are the resources I came up with for the extensions api group:
```
DaemonSet
Deployment
DeploymentRollback
Ingress
NetworkPolicy
PodSecurityPolicy
ReplicaSet
Scale
```
I might go try to figure out what ipamd change sparked the new resource requirement from that pr.

shaunofneuron: @Kishorb @mogren I wasn't able to figure out what specific resources inside extension would be required outside of DaemonSet though I do know there are deprecation changes post 1.16 to the extensions group. I see my changes proposed here as orthogonal to the granularity of clusterrole policies as they represent upstream definitions more closely related to the vpc cni project.  I do however agree with the least privileged approach and think this should be figured out rather soon. I'm not sure what other help I can provide, please let me know!

mogren: @Kishorb I agree that we should scope this down, but I guess for now it would be ok that the Helm chart is matching the current CNI config sample. That said, I opened https://github.com/aws/amazon-vpc-cni-k8s/issues/1232 to address this in the CNI repo, and we should follow up with a PR here to match.

kishorj: sounds good.

kishorj (approved):

PR #258, opened 2020-09-23
rename aws-loadbalancer-controller, update CRD, add securityContext

kishorj:
Issue #, if available:

Description of changes:
Rename to aws-load-balancer-controller
Update rbac, crd, deployment templates
add securityContext for group ID 65534


fawadkhaliq:
Looks good. 

The PR title is misleading as we're also changing security context group ID and updating CRDs. Generally better to have separate PRs or in this case, maybe change the title to something like: "Change aws-loadbalancer-controller name, fsGroup to 65534 and make NetworkingPeer required"

What's the reason for changing fsGroup to 65534 ?

kishorj:
> Looks good.
> 
> The PR title is misleading as we're also changing security context group ID and updating CRDs. Generally better to have separate PRs or in this case, maybe change the title to something like: "Change aws-loadbalancer-controller name, fsGroup to 65534 and make NetworkingPeer required"
> 
> What's the reason for changing fsGroup to 65534 ?
The new docker container is built to run as non root user, so the ﻿﻿﻿secrets mounts are not accessible. The fsGroup in the security context is to make the mounts accessible to this gid.

fawadkhaliq:
> > Looks good.
> > The PR title is misleading as we're also changing security context group ID and updating CRDs. Generally better to have separate PRs or in this case, maybe change the title to something like: "Change aws-loadbalancer-controller name,
> 
> I should've split into multiple commits.
> 
> > fsGroup to 65534 and make NetworkingPeer required"
> > What's the reason for changing fsGroup to 65534 ?
> 
> The new docker container is built to run as non root user, so the ﻿﻿﻿secrets mounts are not accessible. The fsGroup in the security context is to make the mounts accessible to this GID.

Thanks for addressing the comments and clarifying

Issue #177, opened 2020-06-17
Calico Helm chart doesn't support nodeAffinity or nodeSelector configuration


szpuni:
Currently calico helm chart do not allow for nodeSelector or nodeAffinity manipulation. Tolerations are only available in one of deployments are are hardcoded same way as 2 previous attributes.

Above makes impossible to choose nodes in cluster on which calico is installed.

Deployment.yaml file needs to be updated to allow using custom values in nodeSelector, nodeAffinity, tollerations:

```
    spec:
      priorityClassName: system-cluster-critical
      nodeSelector:
        beta.kubernetes.io/os: linux
      tolerations:
        # Mark the pod as a critical add-on for rescheduling.
        - key: CriticalAddonsOnly
          operator: Exists
```

limed:
Resolved in #260

haouc:
@szpuni if you still have concerns, please be aware of our latest announcement. Since we are deprecating Calico charts in this repo, we recommend you to check our [important announcement](https://github.com/aws/eks-charts/tree/master/stable/aws-calico) and suggested [installation instruction](https://docs.aws.amazon.com/eks/latest/userguide/calico.html) for required or latest Calico version. You can also refer to [Calico configuration reference](https://projectcalico.docs.tigera.io/reference/installation/api).
If you need migrate Calico from non-operator based installation to operator based installation, we recommend you to check [this Tigera document](https://projectcalico.docs.tigera.io/maintenance/operator-migration). 
If you need upgrade you calico operator but you didn't install Calico via helm, you can check [this Tigera document](https://projectcalico.docs.tigera.io/getting-started/kubernetes/quickstart). Please being aware that any customized installation may encounter compatibility issue during upgrading.
If you have any issues or questions during Calico resources installation/upgrading/deletion, please refer to the [README](https://github.com/aws/eks-charts/tree/master/stable/aws-calico) and send request to Calico repositories.
I am closing this ticket due to unsupported charts upgrade. Please feel free to send questions to our triage email ([k8s-awscni-triage@amazon.com](mailto:k8s-awscni-triage@amazon.com)).

PR #260, opened 2020-09-24
Don't schedule pods in fargate nodes

limed:
Issue #177 

Description of changes:
This resolves an issue where the aws calico pods tries to get scheduled on a fargate pod, adding affinity to make sure that it will not be scheduled on fargate.

This solution is taken from a similar problem with the `aws-node-termination-handler` chart in commit 86c6e24eab7e22f5edfda7884a827c5119d2c2c4 so if there is a different method of solving this let me know

### DIFF
--- a/stable/aws-calico/templates/daemon-set.yaml
+++ b/stable/aws-calico/templates/daemon-set.yaml
@@ -19,6 +19,15 @@ spec:
         app.kubernetes.io/name: "{{ include "aws-calico.fullname" . }}-node"
     spec:
       priorityClassName: system-node-critical
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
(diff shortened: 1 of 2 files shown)


stable/aws-calico/templates/daemon-set.yaml
```diff
@@ -19,6 +19,15 @@ spec:
...
       priorityClassName: system-node-critical
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                - key: "eks.amazonaws.com/compute-type"
+                  operator: NotIn
+                  values:
+                    - fargate
```
mogren, line 30: Thanks, looks good. We should probably add checks for OS and Arch as well, since the default Calico image will only work on Linux on amd64. That can be a separate PR.

mogren (approved):
🚀

jaypipes (approved):
:+1: awesome, thank you @limed! :)

PR #274, opened 2020-10-08
populate default securityContext

kishorj:
populate default securityContext
add disclaimer for automatic updates

Issue #, if available:

Description of changes:

### DIFF
--- a/stable/aws-load-balancer-controller/values.yaml
+++ b/stable/aws-load-balancer-controller/values.yaml
@@ -38,8 +38,6 @@ securityContext:
   #   - ALL
   readOnlyRootFilesystem: true
   runAsNonRoot: true
-  runAsUser: 1002
-  runAsGroup: 65534
   allowPrivilegeEscalation: false
 
 # Time period for the controller pod to do a graceful shutdown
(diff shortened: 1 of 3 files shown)


stable/aws-load-balancer-controller/values.yaml
```diff
@@ -32,13 +32,15 @@ rbac:
...
+securityContext:
   # capabilities:
   #   drop:
   #   - ALL
-  # readOnlyRootFilesystem: true
-  # runAsNonRoot: true
-  # runAsUser: 1000
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  runAsUser: 1002
```
M00nF1sh, line 41: let's leave runAsUser and runAsGroup unset. (so the docker file controls it).
we only need runAsNonRoot: true to enforce we are not running as root.

M00nF1sh (changes requested):

stable/aws-load-balancer-controller/values.yaml
kishorj: sounds good

M00nF1sh (approved):
lgtm

PR #275, opened 2020-10-11
feat: Add chart for cloudwatch-metrics

gazal-k:
Helm chart created from reference here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-setup-metrics.html

Issue #, if available:

Description of changes:

### DIFF
--- a/stable/aws-cloudwatch-metrics/README.md
+++ b/stable/aws-cloudwatch-metrics/README.md
@@ -1,0 +1,30 @@
+# aws-cloudwatch-metrics
+
+A helm chart for CloudWatch Agent to Collect Cluster Metrics
+
+## Installing the Chart
+
+Add the EKS repository to Helm:
+
+```sh
+helm repo add eks https://aws.github.io/eks-charts
+```
+
+Install or upgrading aws-cloudwatch-metrics chart with default configuration:
+
+```sh
+helm upgrade --install aws-cloudwatch-metrics \
+    --namespace amazon-cloudwatch eks/aws-cloudwatch-metric \
+    --set clusterName=my-eks-cluster
+```
+
+## Configuration
+
+| Parameter | Description | Default | Required |
+| - | - | - | -
+| `image.repository` | Image to deploy | `amazon/cloudwatch-agent` | ✔
+| `image.tag` | Image tag to deploy | `1.247345.36b249270`
+| `image.pullPolicy` | Pull policy for the image | `IfNotPresent` | ✔
+| `clusterName` | Name of your cluster | `cluster_name` | ✔
+| `serviceAccount.create` | Whether a new service account should be created | `true` | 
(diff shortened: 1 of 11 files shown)


stable/aws-cloudwatch-metrics/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Configuration
+
+| Parameter | Description | Default | Required |
+| - | - | - | -
+| `image.repository` | Image to deploy | `amazon/cloudwatch-agent` | ✔
+| `image.tag` | Image tag to deploy | `1.247345.36b249270`
+| `image.pullPolicy` | Pull policy for the image | `IfNotPresent` | ✔
+| `clusterName` | Name of your cluster | `cluster_name` | ✔
+| `serviceAccount.create` | Whether a new service account should be created | `true` |
```
jaypipes, line 29: There is also the `serviceAccount.name` field.

jaypipes (changes requested):
@gazal-k looking really great, thank you! just add the one missing configuration field for service account name and we're good to go.

stable/aws-cloudwatch-metrics/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+{{- end -}}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "aws-cloudwatch-metrics.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
+    {{ default (include "aws-cloudwatch-metrics.fullname" .) .Values.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.serviceAccount.name }}
```
jaypipes, line 60: And one other issue I see is that this is copying in a mistake that was fixed here: https://github.com/aws/eks-charts/commit/8c0db8a9e3845229f4d8806ccc498f2635ac8a29

Basically, you need to remove the `{{ if ` block around the `Values.serviceAccount.create` check above...

jaypipes (changes requested):
One other thing I caught... please see inline :)

stable/aws-cloudwatch-metrics/templates/_helpers.tpl
gazal-k: oops :slightly_smiling_face:

jaypipes (approved):
Awesome work on this @gazal-k thank you! :)

stable/aws-cloudwatch-metrics/templates/_helpers.tpl
jfoechsler: Hi. this service account name template is not used in your cluster role binding. There it just uses  {{ include "aws-cloudwatch-metrics.fullname" . }}. It should also use the template aws-cloudwatch-metrics.serviceAccountName in role binding I think.

PR #283, opened 2020-10-14
Fix irsa fluent bit

allamand:
Issue #, if available:

Description of changes:

When deploying the aws-fluent-bit chart using IRSA (Iam role for Service Account) we generally create the service account appart so that it can include the appropriate role. But we also need the chart to create the proper RBAC needed for the chart to run properly.

I added a new value `rbac.enable` which is true by default and will be test instead of `serviceaccount.enable` when creating RBAC.

it is used with following values.yaml extract:

```yaml
rbac:
  enable: true
serviceAccount:
  create: false
  name: aws-for-fluent-bit
```

### DIFF
--- a/stable/aws-for-fluent-bit/templates/clusterrole.yaml
+++ b/stable/aws-for-fluent-bit/templates/clusterrole.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.serviceAccount.create -}}
+{{- if .Values.rbac.create -}}
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
 metadata:
--- a/stable/aws-for-fluent-bit/templates/clusterrole.yaml
+++ b/stable/aws-for-fluent-bit/templates/clusterrole.yaml
@@ -1,4 +1,3 @@
-{{- if .Values.rbac.create -}}
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
 metadata:
@@ -14,4 +13,3 @@ rules:
     verbs: ["use"]
     resourceNames:
       - {{ include "aws-for-fluent-bit.fullname" . }}
-{{- end -}}
--- a/stable/aws-for-fluent-bit/templates/configmap.yaml
+++ b/stable/aws-for-fluent-bit/templates/configmap.yaml
@@ -158,7 +158,7 @@ data:
         {{- if .Values.elasticsearch.tls }}
         TLS             {{ .Values.elasticsearch.tls }}
         {{- end }}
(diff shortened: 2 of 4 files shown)


stable/aws-for-fluent-bit/templates/configmap.yaml
```diff
@@ -158,7 +158,7 @@ data:
         {{- if .Values.elasticsearch.tls }}
         TLS             {{ .Values.elasticsearch.tls }}
         {{- end }}
-        {{- if .Values.elasticsearch.retry_limit }}
+        {{- if .Values.elasticsearch.retryLimit }}
```
jaypipes, line 161: unrelated change, yes?

stable/aws-for-fluent-bit/templates/clusterrole.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.serviceAccount.create -}}
+{{- if .Values.rbac.create -}}
```
jaypipes, line 1: I would actually just remove the `{{ if }}` guard entirely from the `clusterrole.yaml` and `clusterrolebinding.yaml` templates. None of the other charts have this guard on either their `clusterrole[binding].yaml` or combined `rbac.yaml` template files, FWICT.

stable/aws-for-fluent-bit/values.yaml
```diff
@@ -109,6 +109,8 @@ elasticsearch:
 #     Format template
 #     Template {time} used={Mem.used} free={Mem.free} total={Mem.total}
 
+rbac:
+  create: true
```
jaypipes, line 113: And just remove this new configuration option...

jaypipes (changes requested):
I think I'd rather just remove the if guards entirely...

stable/aws-for-fluent-bit/templates/configmap.yaml
allamand: I correct an issue with my last or also

stable/aws-for-fluent-bit/templates/clusterrole.yaml
allamand: Yes agree. I'll entirely remove them

stable/aws-for-fluent-bit/values.yaml
allamand: Yes

allamand (commented):
@jaypipes take into account your suggestion

jaypipes (approved):
:+1: sorry for the delay on this @allamand!

PR #299, opened 2020-10-23
[aws-load-balancer-controller]Add extraArgs

cw-sakamoto:
Signed-off-by: cw-sakamoto <sakamoto@chatwork.com>

Issue #, if available:

Description of changes:
- I'd like to set some arguments to the command args, but I think it depends on the use case, so I've made it so that anything can be set as `extraArgs`

### DIFF
--- a/stable/aws-load-balancer-controller/Chart.yaml
+++ b/stable/aws-load-balancer-controller/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-load-balancer-controller
 description: AWS Load Balancer Controller Helm chart for Kubernetes
-version: 1.0.1
+version: 1.0.2
 appVersion: v2.0.0
 home: https://github.com/aws/eks-charts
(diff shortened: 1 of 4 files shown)


cw-sakamoto:
@kishorj Thank you for your this chart maintenance. 
Can you review it?

dza89:
I think this is better:
https://github.com/aws/eks-charts/pull/300

kishorj:
Thank you for your contribution. We definitely need more configuration options, but IMO, we should add them individually. The blanket extraArgs is a powerful construct, however it is also prone to errors. Also, there is no documentation in the chart for what values can be safely used, and would require the user to figure them out from different places.

If there are specific configuration option that are useful for your use-case, please feel free to open a PR.

jaypipes (commented):
Agree with @kishorj. The pattern we have used in all the other Charts is to use specific configuration options matching specific CLI arguments.

kishorj:
I'm closing this PR since we do not prefer using the extraArgs option as proposed. Please feel free to open new PRs for specific config options.

PR #301, opened 2020-10-25
[aws-load-balancer-controller] Fix RBAC permissions on leader election role

kishorj:
Issue #, if available:

Description of changes:
Undo PR #298 changes since we cannot restrict create request by resourceName. For create, this limitation is because the object name is not known at authorization time. 

Tests
- Installed chart without existing configmap aws-load-balancer-controller-leader in kube-system. Verified chart installed successfully and leader election worked.
- Installed chart with existing configmap aws-load-balancer-controller-leader

For further reference https://kubernetes.io/docs/reference/access-authn-authz/rbac/

### DIFF
--- a/stable/aws-load-balancer-controller/templates/rbac.yaml
+++ b/stable/aws-load-balancer-controller/templates/rbac.yaml
@@ -8,8 +8,11 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: [configmaps]
+  verbs: [create]
+- apiGroups: [""]
+  resources: [configmaps]
   resourceNames: [aws-load-balancer-controller-leader]
-  verbs: [create, get, patch, update]
+  verbs: [get, patch, update]
 ---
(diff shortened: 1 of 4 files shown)


stable/aws-load-balancer-controller/templates/rbac.yaml
```diff
@@ -8,7 +8,6 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: [configmaps]
-  resourceNames: [aws-load-balancer-controller-leader]
```
fawadkhaliq, line 11: See my comment here please: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1554/files#r511543948

kishorj: sure.

fawadkhaliq (approved):
/lgtm

PR #312, opened 2020-10-28
[aws-load-balancer-controller]Add aws-load-balancer-controller args

cw-sakamoto:
Issue #, if available:

Description of changes:
- Referring to [configurations](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/controller/configurations.md), I've added the args we use in v1 and the args we will use in the future.

### DIFF
--- a/stable/aws-load-balancer-controller/templates/deployment.yaml
+++ b/stable/aws-load-balancer-controller/templates/deployment.yaml
@@ -75,6 +75,9 @@ spec:
         {{- if .Values.serviceMaxConcurrentReconciles }}
         - --service-max-concurrent-reconciles={{ .Values.serviceMaxConcurrentReconciles }}
         {{- end }}
+        {{- if .Values.targetgroupbindingMaxConcurrentReconciles }}
+        - --targetgroupbinding-max-concurrent-reconciles={{ .Values.targetgroupbindingMaxConcurrentReconciles }}
+        {{- end }}
         {{- if .Values.logLevel }}
         - --log-level={{ .Values.logLevel }}
         {{- end }}
--- a/stable/aws-load-balancer-controller/templates/deployment.yaml
+++ b/stable/aws-load-balancer-controller/templates/deployment.yaml
@@ -63,9 +63,6 @@ spec:
         {{- if .Values.enableWafv2 }}
         - --enable-wafv2={{ .Values.enableWafv2 }}
         {{- end }}
-        {{- if .Values.healthProbeBindAddr }}
-        - --health-probe-bind-addr={{ .Values.healthProbeBindAddr }}
-        {{- end }}
         {{- if .Values.metricsBindAddr }}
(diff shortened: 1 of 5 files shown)


stable/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -60,15 +99,22 @@ spec:
...
         - name: webhook-server
-          containerPort: 9443
+          containerPort: {{ .Values.webhookBindPort | default 9443 }}
           protocol: TCP
         - name: metrics-server
-          containerPort: 8080
+          containerPort: {{ (split ":" .Values.metricsBindAddr)._1 | default 8080 }}
+          protocol: TCP
+        {{- if .Values.livenessProbe }}
+        - name: health
```
kishorj, line 108: any reason this port needs to be exposed?

stable/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -48,6 +48,45 @@ spec:
...
+        - --health-probe-bind-addr={{ .Values.healthProbeBindAddr }}
+        {{- end }}
+        {{- if .Values.metricsBindAddr }}
+        - --metrics-bind-addr={{ .Values.metricsBindAddr }}
+        {{- end }}
+        {{- if .Values.ingressMaxConcurrentReconciles }}
+        - --ingress-max-concurrent-reconciles={{ .Values.ingressMaxConcurrentReconciles }}
+        {{- end }}
+        {{- if .Values.serviceMaxConcurrentReconciles }}
+        - --service-max-concurrent-reconciles={{ .Values.serviceMaxConcurrentReconciles }}
```
kishorj, line 76: Since we are at it, lets also add an option for targetgroupbinding-max-concurrent-reconciles

stable/aws-load-balancer-controller/templates/deployment.yaml
cw-sakamoto: The reason is that I wanted to use this name in the livenessProbe port.

When the healthcheck port is can be changed by this PR(`healthProbeBindAddr`), the livenessProbe port also needs to be changed, but I wanted to give it a name so that no changes are missed.

https://github.com/aws/eks-charts/pull/312/files#diff-16853b1bc086bb28c85a18bece05525988df3c4d6caeb4d12e3596979fa3a996R129

cw-sakamoto: I understand. I'll just fix the part about `health-probe-bind-addr` for now.

cw-sakamoto: It isn't arg(health-probe-bind-addr) I wanted to use right away, so I deleted the arg and rebased.
ab79529
f709626

kishorj (approved):
Thanks for your effort, I appreciate it.

PR #314, opened 2020-10-29
[aws-load-balancer-controller] Add aws load balancer controller env

8398a7:
Issue #, if available:

Description of changes:

The previous chart allowed for the injection of extra values.
https://github.com/helm/charts/tree/master/incubator/aws-alb-ingress-controller

I think these would be better if they were available because it's hard to keep track of the chart side every time a new flag is added.
For example, I used it to disable waf.

The environment variables were supported before, and I hope they will be supported this time too.


kishorj:
We'd prefer to enable the flags individually. If the flags enabled from the chart are known, we can keep them up to date with the changes to the controller.

Please check if the recent changes from PR https://github.com/aws/eks-charts/pull/312 help your use case.

stable/aws-load-balancer-controller/values.yaml
```diff
@@ -87,3 +87,9 @@ livenessProbe:
     scheme: HTTP
   initialDelaySeconds: 30
   timeoutSeconds: 10
+
+extraArgs: {}
+
+extraEnv: {}
```
kishorj, line 93: Lets name it env

stable/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -48,6 +48,14 @@ spec:
         {{- if .Values.vpcId }}
         - --aws-vpc-id={{ .Values.vpcId }}
         {{- end }}
+        {{- range $key, $value := .Values.extraArgs }}
+          - --{{ $key }}={{ $value }}
+        {{- end }}
+        env:
```
kishorj, line 54: check if env is non-empty
{{- if .Values.env }}

8398a7:
Thank you. I see the recent PR is closed #299 as well.
I'm going to re-issue a PR that only adds environment variables since the WAF feature is in #312.

stable/aws-load-balancer-controller/values.yaml
kishorj, line 93: add comment similar to - 
Environment variables to set for aws-load-balancer-controller pod

kishorj (commented):
Env looks useful

kishorj:
> Thank you. I see the recent PR is closed #299 as well.
> I'm going to re-issue a PR that only adds environment variables since the WAF feature is in #312.

You can update this PR with the env changes

8398a7:
I'm sorry. I'm getting ahead of myself. I'll just fix the one about env in this PR.

stable/aws-load-balancer-controller/values.yaml
8398a7: Fixed :+1:

stable/aws-load-balancer-controller/templates/deployment.yaml
8398a7: Fixed :+1:

stable/aws-load-balancer-controller/values.yaml
8398a7: Fixed :+1:

kishorj (approved):

dza89:
How is this ever approved, who puts their credentials in your environment, even as an example.

kishorj:
> How is this ever approved, who puts their credentials in your environment, even as an example.

@dza89 Thanks for highlighting this issue. You concern is valid, putting credentials even as an example has the potential to legitimize bad security practices. I've updated the comments, PR #326.

Issue #277, opened 2020-10-12
Why is aws-for-fluent-bit memory request 500Mi?


andrew-e:
Hello,

Out of the box fluent-bit memory request 8Mi per pod, and can probably run on lower, since the point of fluent-bit was to be a low memory solution.  Yet, the values.yaml defaults to 500Mi.  Why is this?  Are you adding 450 mb of dependencies?

Apoogies that this isn't a bug per se, but I am curious to know if I really need to allocate a sizable portion of memory for each node for a log aggregator.

I pulled and ran the Docker image locally, and in an idle state, it has these metrics...
```
CONTAINER ID        NAME                 CPU %               MEM USAGE / LIMIT    MEM %               NET I/O             BLOCK I/O           PIDS
291cb539e184        optimistic_mestorf   0.01%               25.7MiB / 4.826GiB   0.52%               836B / 0B           0B / 0B             21
```

So, both memory and cpu requests seem excessive.

PettitWesley:
There's an open PR to fix this. The request is too high. As you have seen, in most cases memory usage is quite low.

dekarl:
I think this issue has served its purpose as #323 has since been merged. 🎉

PR #323, opened 2020-11-02
[aws-for-fluentbit] Reduce resource limits/requests

mickeypash:
I encountered the issue when deploying to a smaller development cluster.

According to Eduardo Silva and Wesley Pettit people in production mostly
have 150MB as a max limit in terms of memory.

Timestamped to the answer of the question:
"What should requests/limits be set to for production?"

- [Advanced Logging with Fluent Bit - Eduardo Silva, Arm Treasure Data & Wesley Pettit, Amazon](https://www.youtube.com/watch?v=F73MgV_c2MM&feature=youtu.be&t=2200)

Tries to address #277


mickeypash:
A slightly different set of defaults is set as part of the AWS Container Insights Demo [here for the Fluentd agent](https://github.com/aws-samples/amazon-cloudwatch-container-insights/blob/master/k8s-deployment-manifest-templates/deployment-mode/daemonset/container-insights-monitoring/fluentd/fluentd.yaml#L397-L400)
```
          resources:
            limits:
              memory: 400Mi
            requests:
              cpu: 100m
              memory: 200Mi
```

Gowiem:
Just chiming into mention that this should be merged / addressed, the current defaults are excessively high when the whole purpose of Fluent Bit is to be very minimal from the CPU / Memory perspective. I'll be using @mickeypash's values from this PR for my own purposes until it is.

stable/aws-for-fluent-bit/values.yaml
```diff
@@ -116,10 +116,10 @@ serviceAccount:
 
 resources:
   limits:
-    memory: 500Mi
+    memory: 150Mi
```
PettitWesley, line 119: I strongly agree with lowering the request. Lowering the limit is more dangerous. Putting it at 250 would be safer for more folks. If you have a lot of logs it can and will use that much memory.

The problem this PR wanted to address was the inability to run this on a small instance- and changing the request should fix that right? but the limit could be kept higher? Or am I talking nonsense.

PettitWesley (approved):
There's no perfect value to set that will make everyone happy. I think this is sufficient for a lot of users.

stable/aws-for-fluent-bit/values.yaml
mickeypash: Hey Wesley!

PettitWesley: Cool! In retrospect I wish I could amend my answer to that question a little bit. It's not abnormal to see memory usage up to a few hundred MB if you are sending a lot of logs or doing a lot of processing. And I think my estimate for the go plugins was a bit off, you can easily see pretty very high memory usage with those. The answer I gave is accurate for the most common cases with the core plugins.

mickeypash: Thanks! I've amended the `limits` to now be 250Mi 👍

mickeypash:
Just wanted to follow up.
Wondering if there's anything else I need to do to get this PR closed?

PR #335, opened 2020-11-09
appmesh-controller: bump version to v1.2.0 and add note on iamserviceaccount create

fawadkhaliq:
**Description of changes**
aws-app-mesh-controller-for-k8s has [released version 1.2.0](https://github.com/aws/aws-app-mesh-controller-for-k8s/releases/tag/v1.2.0). Bumping the Helm chart to use the v1.2.0. Also, added a note on iamserviceaccount override in the README

### DIFF
--- a/stable/appmesh-controller/README.md
+++ b/stable/appmesh-controller/README.md
@@ -62,8 +62,10 @@ aws iam create-policy \
 ```
 Take note of the policy ARN that is returned
 
-
 Create an IAM role for service account for the App Mesh Kubernetes controller, use the ARN from the step above
+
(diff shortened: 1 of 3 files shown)


stable/appmesh-controller/README.md
```diff
@@ -62,8 +62,10 @@ aws iam create-policy \
 ```
 Take note of the policy ARN that is returned
 
-
 Create an IAM role for service account for the App Mesh Kubernetes controller, use the ARN from the step above
+
+> Note: if you deleted `serviceaccount` in the `appmesh-system` namespace, you will need to delete and re-create `iamserviceaccount`. `eksctl` does not override the `iamserviceaccount` correctly ([see this issue](https://github.com/weaveworks/eksctl/issues/2665))
```
kishorj, line 67: It would be less confusing if we could show how to delete existing iamserviceaccount.

fawadkhaliq: I did not add that on purpose as some users run commands copy-pasta style _without_ reading the comments so they may run the delete command without reading the reason why it needs to be run

kishorj (approved):

PR #343, opened 2020-11-12
[aws-load-balancer-controller] Add aws load balancer controller election settings

s2504s:
#### What:
* Add the ability to set a name of namespace and name of configmap for a leader elections procedure

#### Why:
* When we use more that one controller we should set these options in order to avoid using the same name of the ConfigMap

Description of changes:


stable/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -90,6 +90,12 @@ spec:
         {{- if .Values.watchNamespace }}
         - --watch-namespace={{ .Values.watchNamespace }}
         {{- end }}
+        {{- if index .Values "election-id" }}
+          - --election-id={{ .Values index "election-id" }}
```
fawadkhaliq, line 94: You will need to update the resource name in the rbac as well here [1] 

[1] https://github.com/aws/eks-charts/blob/master/stable/aws-load-balancer-controller/templates/rbac.yaml#L14

s2504s: I have added the condition for using this value in RBAC rules if that value has been set

kishorj:
Could you provide more details on your use case of running multiple controllers in the same cluster with different leader election ids? As it stands currently, there will be issues with the NLB IP feature if multiple controllers are run.

stable/aws-load-balancer-controller/README.md
```diff
@@ -161,3 +161,5 @@ The default values set by the application itself can be confirmed [here](https:/
 | `livenessProbe`                             | Liveness probe settings for the controller                                                               | (see `values.yaml`)                                                                |
 | `env`                                       | Environment variables to set for aws-load-balancer-controller pod                                        | None                                                                               |
 | `hostNetwork`                               | If `true`, use hostNetwork                                                                               | `false`                                                                            |
+|`leader-election-namespace`                  | Namespace of leader-election configmap                                                                   |(see `values.yaml`)                                                                 |
```
kishorj, line 164: Lets use camelCase to be consistent with other parameters

stable/aws-load-balancer-controller/templates/rbac.yaml
```diff
@@ -51,6 +51,10 @@ rules:
 - apiGroups: [""]
   resources: [nodes, secrets, namespaces, endpoints]
   verbs: [get, list, watch]
+- apiGroups: [""]
+  resources: [configmaps]
```
kishorj, line 55: Any reason this permission is required?

stable/aws-load-balancer-controller/values.yaml
```diff
@@ -117,6 +117,12 @@ syncPeriod:
 # Namespace the controller watches for updates to Kubernetes objects, If empty, all namespaces are watched.
 watchNamespace:
 
+# Namespace of leader-election configmap for ingress controller. If unspecified, the namespace of this controller pod will be used
```
kishorj, line 120: Lets refer as controller or the load balancer controller

kishorj:
@s2504s any updates?

kishorj:
Closing the pull request since we don't intend to support running multiple controller deployments in the same cluster.

PR #346, opened 2020-11-13
[aws-load-balancer-controller] Add extraVolumes and extraVolumeMounts

utay:
Adding `extraVolumes` and `extraVolumeMounts` to the chart so that we can manually setup IRSA on our non-EKS Kubernetes cluster. 

Docs:
- https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/
> If you do not want to the admission controller to modify your pods, you can manually add the environment variables `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` with the values of a projected service account token location and the role to assume. In addition, you will also need to add a `volume` and `volumeMounts` parameter to the pod with a projected service account token, see the Kubernetes docs for reference.
- https://kubernetes.io/docs/concepts/storage/volumes/#projected


utay:
@kishorj Any chance you have time to look at this? Thanks!

kishorj (approved):

PR #351, opened 2020-11-16
Support the managedAsgTag option in aws-node-termination-handler

blakestoddard:
Adds support for the configurable `managedAsgTag` option in aws-node-termination-handler.

NTH PR: https://github.com/aws/aws-node-termination-handler/pull/287


blakestoddard:
Whoops, this belongs in the aws-node-termination-handler repo, not here.

PR #373, opened 2020-11-25
[aws-load-balancer-controller] Update apiVersion for MutatingWebhookConfiguration because of deprecation in Kubernetes v1.19

dieterdemeyer:
Description of changes:

Updates `apiVersion` for `MutatingWebhookConfiguration` files because `v1beta1` can no longer be used in Kubernetes clusters with version `v1.19` and onwards.


kishorj:
We use v1beta1 version for backwards compatibility with older EKS versions. I did test the chart on kops clusters built from the k8s master. Did you run into any specific error with the chart?

dieterdemeyer:
@kishorj , I didn't run into specific problems yet. But when running the tool pluto against our running cluster, the aws-load-balancer-controller came up as using a deprecated apiVersion for the MutatingWebhookConfiguration and that it would be removed starting from Kubernetes v.1.9.0 (cfr. API reference [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#mutatingwebhookconfiguration-v1-admissionregistration-k8s-io))
So I was under the assumption that the helm chart would fail when being deployed in a cluster with that version onwards.

Perhaps another solution would be to be able to specify the `apiVersion` in the `values.yaml` file ?
That way, it's configurable and will still be backwards compatible.

kishorj:
We will defer this change until EKS 1.15 is EOL.

PR #382, opened 2020-12-08
[aws-for-fluent-bit]Add Buffer_Size parameter to Kubernetes Filter

cw-sakamoto:
Signed-off-by: cw-sakamoto <sakamoto@chatwork.com>

Issue #, if available:
https://github.com/fluent/fluent-bit/issues/2111

Description of changes:
`managedFileds`, introduced in 1.18, causes the response size from API server to increase, which can cause buffer to overflow and no metadata to enter.

In `values.yaml` it defaults to 32k and can be specified as a parameter.


cw-sakamoto:
@jaypipes 
Thanks for this chart maintenance. We've troubled this problem...It would be helpful if you could review it sooner.

stable/aws-for-fluent-bit/values.yaml
```diff
@@ -43,6 +43,7 @@ filter:
   mergeLogKey: "data"
   k8sLoggingParser: "On"
   k8sLoggingExclude: "On"
+  bufferSize: "32k"
```
PettitWesley, line 46: Just noting that 32KB is the current default

PettitWesley (approved):
So basically this makes the parameter configurable for a user?

marklinders-planon:
Why would we not just set the Buffer_Size to 0? Is there a big downside to running "unlimited?"

PettitWesley:
@marklinders-planon I agree; I don't see any downside to setting it to unlimited mode (0).

BartXZX:
Makes me wonder why this is configurable at all, if users are going to set it to 0 anyway without any downside. 🤔

PettitWesley:
May be lets update the PR to set it to 0 instead of making it configurable?

BartXZX:
Sorry I'm kind of hijacking the thread; we're not using the Chart ourselves, and were just looking into the issue @marklinders-planon mentioned. If there's no clear benefit, maybe remove the setting from FluentBit altogether?

PettitWesley:
Questions about why `Buffer_Size` was even created in the first place should be directed to @edsiper

edsiper:
control memory usage for large response scenarios.

PettitWesley:
how big would the response for a request to get metadata be, realistically? If the Fluent Bit pod is given 100 MB limit for example, I doubt it will be a significant fraction of that. Implying- I don't think kubernetes users need to care about tuning memory usage via this parameter.

edsiper:
yeah, I would trust Kubernetes end-point; but for example, some cases we cannot trust Elastic responses

stable/aws-for-fluent-bit/values.yaml
cw-sakamoto: @PettitWesley 
Thanks for your comment. The following document says 32k, is that wrong?
https://docs.fluentbit.io/manual/pipeline/filters/kubernetes

cw-sakamoto:
>how big would the response for a request to get metadata be, realistically? If the Fluent Bit pod is given 100 MB limit for example, I doubt it will be a significant fraction of that. Implying- I don't think kubernetes users need to care about tuning memory usage via this parameter.

I don't know why 32k is the [default](https://docs.fluentbit.io/manual/pipeline/filters/kubernetes), but as far as the filter implementation is concerned, it is just a setting for the size of the buffer to receive responses from api-server, so unless you have a very small device (e.g. embedded device), I don't think I will have any problems with 0.
Shall I fix it at 0?

tpoindessous:
Hi ! Any update on this change ? 

thanks !

cw-sakamoto:
@tpoindessous Conflict resolved. Please check this PR.

stable/aws-for-fluent-bit/values.yaml
PettitWesley: No it isn't wrong... I think I might have made this comment since I wasn't sure if we should just copy the same default or create a different one here

PettitWesley (approved):
@cw-sakamoto This PR can not be merged yet because the CI failed. That needs to pass.

cw-sakamoto:
@PettitWesley Thank you for your comment. I updated a file and the CI passed.

PR #383, opened 2020-12-10
Delete invalid status definition

unaiwillr:
Keeping this here breaks pulumi deployments as `status` is not a valid attribute of `CustomResourceDefinition`.

Issue #, if available: n/a

Description of changes:
Delete an invalid attribute, `status` on `CustomResourceDefintion`


kishorj:
Could you elaborate more on why `status` is an invalid attribute?

The CRD is auto-generated by kubebuilder, so we don't prefer to make changes in this repo. Any changes must be made in the controller repo https://github.com/kubernetes-sigs/aws-load-balancer-controller

PR #384, opened 2020-12-10
aws-vpc-cni: move default affinity to values.yaml

sarahhodne:
This change allows you to override the affinity, should you want to, using the already existing `affinity` key in the Helm values.

Previously, setting the `affinity` key in the Helm values would cause a duplicate `affinity` key in the pod spec. This might have worked, but I think it would be dependent on the YAML parser


jayanthvn (approved):
LGTM, Thanks.

PR #397, opened 2020-12-18
[aws-cloudwatch-metrics] add the possibility to change the hostNetwork

frankyhun:
Issue #, if available:
https://www.reddit.com/r/aws/comments/g2huur/deploying_cloudwatch_agent_in_eks_as_a_pod_is/

Description of changes:
Add the possibility to change the hostNetwork via passing it trough the values. 
Set default to false for backwards compatibility.

### DIFF
--- a/stable/aws-cloudwatch-metrics/templates/daemonset.yaml
+++ b/stable/aws-cloudwatch-metrics/templates/daemonset.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     {{- include "aws-cloudwatch-metrics.labels" . | nindent 4 }}
 spec:
-  hostNetwork: {{ .Values.hostNetwork | default false}}
+  hostNetwork: {{ .Values.hostNetwork }}
   selector:
     matchLabels:
       {{- include "aws-cloudwatch-metrics.selectorLabels" . | nindent 6 }}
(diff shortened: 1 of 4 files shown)


stable/aws-cloudwatch-metrics/templates/daemonset.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   labels:
     {{- include "aws-cloudwatch-metrics.labels" . | nindent 4 }}
 spec:
+  hostNetwork: {{ .Values.hostNetwork | default false}}
```
jaypipes, line 8: Please add a `hostNetwork` element to the [`values.yaml`](https://github.com/aws/eks-charts/blob/master/stable/aws-cloudwatch-metrics/values.yaml) file itself.

Simply refer to that value in the daemonset.yaml template here, instead of using a `| default false` statement here.

jaypipes (changes requested):
Hi @frankyhun! Please see inline comment and I request that you also please add an entry for the next variable in the chart's README doc:

https://github.com/aws/eks-charts/blob/master/stable/aws-cloudwatch-metrics/README.md

Thanks!
-jay

stable/aws-cloudwatch-metrics/templates/daemonset.yaml
frankyhun: done

jaypipes (approved):
Great, thank you @frankyhun!

PR #406, opened 2021-01-04
[aws-cloudwatch-metrics] Wrong place of hostNetwork field.

frankyhun:
No issue yet, but the 0.0.2 chart containing my changes https://github.com/aws/eks-charts/pull/397 is unusable, due the hostNetwork field i tried to include in the wrong yaml.

Description of changes:
The hostNetwork field is set in the DaemonSet, but that is wrong. It has to be set in the PodSecurityPolicy, and the PodSecurityPolicy has to be used in the cluster role.

Btw. my previous changes resulted in a WARNING in the CircleCI. Maybe it would be a good idea to break the pipeline in errors like that, see:
https://app.circleci.com/pipelines/github/aws/eks-charts/848/workflows/b97f1314-03e9-4716-819e-7b8f3dbbacea/jobs/2134

WARN - aws-cloudwatch-metrics/templates/daemonset.yaml contains an invalid DaemonSet (release-name-aws-cloudwatch-metrics) - hostNetwork: Additional property hostNetwork is not allowed


jaypipes:
@frankyhun `hostNetwork` in the Daemonset's `spec.template.spec` is allowed (the Pod spec itself, not the Daemonset's spec). Would it be better to move the `hostNetwork` field there instead of using PSP, which a) is in v1beta1 version and b) I'm not sure is a long-term solution compared to something like Gatekeeper/OPA.

jaypipes (approved):
:+1: 

And yes, @frankyhun it would be a very good idea to gate on warnings like this in our CI!

PR #417, opened 2021-01-09
aws-for-fluent-bit: Add the ability to add annotations to the daemonset

georgyo:
Issue #, if available:

Description of changes:
The current chart explicitly requires using the host's metadata credentials to talk to cloudhealth. This means that that you cannot block that metadata access for any of your pods or use a solution like kiam.

This chart requires the ability to add annotations like `iam.amazonaws.com/role` or `eks.amazonaws.com/role-arn` to allow the use of solutions that provide correct credentials to these pods.

Alternatively, the chart could allow using the host networking, but I much rather it support it using the correct role provided by a solution like kiam.


PettitWesley (approved):
Allowing annotations seems like an unobjectionable change

JonathanLachapelle (approved):

PR #480, opened 2021-03-23
Added Instructions for setting up Envoy policies

cgchinmay:
### Description of changes
Added Instructions for Attaching Envoy Policies to the envoy proxy when using IRSA

### Checklist
- Added/modified documentation as required (such as the `README.md` for modified charts)

### Testing
Verified the steps by executing them locally on a cluster

### DIFF
--- a/stable/appmesh-controller/README.md
+++ b/stable/appmesh-controller/README.md
@@ -106,15 +106,10 @@ aws iam create-policy \
 
 Take note of the policy ARN that is returned
 
-Create Application namespace to which Envoy proxy will get injected
-```
-kubectl create ns <ApplicationNamespaceName>
-```
-
-Create an IAM role for service account for the App Mesh Envoy, use the ARN from the step above
+Create an IAM role for service account for the App Mesh Envoy, use the ARN from the step above. Ensure that Application Namespace already exists  
 ```
 eksctl create iamserviceaccount --cluster $CLUSTER_NAME \
-    --namespace <ApplicationNamespaceName> \
+    --namespace <ApplicationNamespaceName to which Envoy gets Injected> \
     --name envoy-proxy \
     --attach-policy-arn arn:aws:iam::$AWS_ACCOUNT_ID:policy/AWSAppMeshEnvoyIAMPolicy  \
     --override-existing-serviceaccounts \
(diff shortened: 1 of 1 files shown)


stable/appmesh-controller/README.md
```diff
@@ -86,11 +86,48 @@ helm upgrade -i appmesh-controller eks/appmesh-controller \
...
+    --approve
+```
+
+Reference this Service Account in your deployment pod spec. This should be the pod which would get injected with the Envoy. Refer below example:
+```
+https://github.com/aws/aws-app-mesh-examples/blob/5a2d04227593d292d52e5e2ca638d808ebed5e70/walkthroughs/howto-k8s-fargate/v1beta2/manifest.yaml.template#L220
+``` 
+
+#### Without IRSA   
+If not setting up IAM role for service account, apply the IAM policies manually to your worker nodes:
```
achevuru, line 130: Do we need to add them when not using IRSA? Did we verify that?

stable/appmesh-controller/README.md
```diff
@@ -86,11 +86,48 @@ helm upgrade -i appmesh-controller eks/appmesh-controller \
...
+Create an IAM policy called AWSAppMeshEnvoyIAMPolicy
+```
+aws iam create-policy \
+    --policy-name AWSAppMeshEnvoyIAMPolicy \
+    --policy-document file://envoy-iam-policy.json
+```
+
+Take note of the policy ARN that is returned
+
+Create Application namespace to which Envoy proxy will get injected
```
achevuru, line 109: I would say we don't need this step. Instead we can directly jump to the next step and ask them to fill in the relevant application namespace(s). We know that a Mesh can be spread across multiple namespaces and so this is required across all the namespaces that the mesh selects (or) their application is running.

stable/appmesh-controller/README.md
cgchinmay: We already had this in the doc.

stable/appmesh-controller/README.md
cgchinmay: Got it. Will update this

cgchinmay:
Closing this pull request. As this should be moved to appmesh-controller repo instead 
https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/458

PR #509, opened 2021-04-20
Added PodDisruptionBudget specification for the appmesh-controller

cgchinmay:
### Description of changes
Add support to specify podDisruptionBudget for the appmesh-controller pods via helm chart 

Added PodDisruptionBudget specification which would allow customers to specify podDisruptionBudget for appmesh-controller pods

### Checklist

### Testing
Schedule 2 appmesh-controller pods with PodDisruptionBudget set 1. Tried draining both nodes. Here is the output
```
error when evicting pods/"appmesh-controller-bf89cd659-8pdrj" -n "appmesh-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
pod/coredns-5946c5d67c-n9wcq evicted
pod/coredns-5946c5d67c-khxvn evicted
evicting pod appmesh-system/appmesh-controller-bf89cd659-8pdrj
error when evicting pods/"appmesh-controller-bf89cd659-8pdrj" -n "appmesh-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```
PDB prevented draining both pods

### DIFF
--- a/stable/appmesh-controller/values.yaml
+++ b/stable/appmesh-controller/values.yaml
@@ -122,6 +122,11 @@ stats:
 # Enable cert-manager
 enableCertManager: false
 
+# podDisruptionBudget for Appmesh controller
+pdb:
+  # minAvailable pods for appmesh-controller
+  minAvailable: 0
+
 # Environment variables to set in appmesh-controller pod
 env: {}
 
--- a/stable/appmesh-controller/Chart.yaml
+++ b/stable/appmesh-controller/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: appmesh-controller
 description: App Mesh controller Helm chart for Kubernetes
-version: 1.3.2
+version: 1.3.3
 appVersion: 1.3.0
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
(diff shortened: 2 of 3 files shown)


stable/appmesh-controller/values.yaml
```diff
@@ -122,6 +122,11 @@ stats:
 # Enable cert-manager
 enableCertManager: false
 
+# podDisruptionBudget for Appmesh controller
+pdb:
+  # minAvailable pods for appmesh-controller
+  minAvailable: 0
```
achevuru, line 128: Why '0'?? Should be at least '1'

achevuru:
Let's close this and raise the PR against appmesh-contoller-repo. My other concern is that I believe `PodDisruptionBudget` is stable only from v1.21 and so I'm not sure we should even include it in the default helm chart that we vend out.

stable/appmesh-controller/values.yaml
cgchinmay: By default there was never any PodDisruptionbudget so had assigned to 0. But, yes I agree, it should be 1 instead so that we have at least 1 pod running.

cgchinmay:
> Let's close this and raise the PR against appmesh-contoller-repo. My other concern is that I believe `PodDisruptionBudget` is stable only from v1.21 and so I'm not sure we should even include it in the default helm chart that we vend out.

I will confirm that.

cgchinmay:
Will open PR against controller repo instead

PR #519, opened 2021-05-12
[aws-load-balancer-controller] release controller v2.2.0

kishorj:
### Issue
New chart for controller version v2.0.0
Includes default HA configuration
- replica count 2 
- controller pods priority class system-cluster-critical
- soft pod anti-affinity set so as to not run two controller pods on the same node 

### Checklist

### Testing
Updated the IAM policy, updated the CRDs and upgraded the chart. Verified that the controller is able to reconcile service/ingress resources as expected.


stable/aws-load-balancer-controller/templates/deployment.yaml
```diff
@@ -135,8 +135,17 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-    {{- with .Values.affinity }}
       affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
M00nF1sh, line 140: i think it's better to be `preferredDuringSchedulingIgnoredDuringExecution`

kishorj: ack

stable/aws-load-balancer-controller/templates/webhook.yaml
```diff
@@ -36,7 +36,7 @@ webhooks:
     - key: app.kubernetes.io/name
       operator: NotIn
       values:
-      - aws-load-balancer-controller
+      - {{ include "aws-load-balancer-controller.fullname" . }}
```
M00nF1sh, line 39: seems we are using `aws-load-balancer-controller.name`

kishorj: thanks. I will update

M00nF1sh (approved):
/lgtm

PR #527, opened 2021-05-24
[aws-node-termination-handler]: Use hostPort only with hostNetwork

hwoarang:
### Description of changes

There is no need to use the hostPort option if we are not directly
connected to the hostNetwork. Using the hostPort option makes that port
unavailable for binding to any other service on the node which is
not really desirable if the pod itself is not on the host network.

### Checklist

### Testing

No testing was done apart from checking with helmTemplate that no `hostPort` is
present in the rendered files when useHostNetwork is not enabled.

### DIFF
--- a/stable/aws-node-termination-handler/Chart.yaml
+++ b/stable/aws-node-termination-handler/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-node-termination-handler
 description: A Helm chart for the AWS Node Termination Handler
-version: 0.15.0
+version: 0.15.1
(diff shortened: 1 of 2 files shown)


bwagner5:
Sorry for the delay! We only take PRs to NTH through the application's repo here https://github.com/aws/aws-node-termination-handler , this allows us to run our suite of e2e tests and then we sync over the chart to eks-charts when we release NTH . Would you mind opening this PR in that repo?  

The changes look good here. When you do open the PR in the NTH repo, you can drop the chart version change since it will automatically be incremented when we sync the chart here. Thanks!

PR #600, opened 2021-09-13
[aws-calico] Updated CRDs to v1 API

jasonumiker:
The API of the CRDs was apiextensions.k8s.io/v1beta1 and was throwing a warning and will be removed in 1.22.

I updated the to the v1 API which required explicitly defining their schema. I didn't come up with the update myself - I got the updated Specs from https://github.com/aws/amazon-vpc-cni-k8s/tree/master/charts/aws-calico/templates/crds

I did put them all together in one file like they were here originally though.


Sir-Jacques:
Review/approval is greatly appreciated!

jdn5126:
Closing as this chart is no longer vendored

PR #644, opened 2021-11-30
Set the namespace for the ALBC PodDisruptionBudget

Dutchy-:
This sets the namespace for the PDB in the aws-load-balancer-controller


kishorj:
@Dutchy-, thanks for your contribution. Would you be able to open a PR in the upstream repository https://github.com/kubernetes-sigs/aws-load-balancer-controller/ instead?

Dutchy-:
Done, https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2391 , thank you

Issue #657, opened 2021-12-13
When using extra input and/or filter and/or output, the configmap created is not correct and fluentbit failed to start


ltupin:
When you want to use extra filter/input/output options it is confusing in comments to leave the title itself. 

## Example 1:

The example you will find in `values.yaml` is:
```
# extraFilters: |
#   [FILTER]
#       Name   grep
#       Match  *
#       Exclude log lvl=debug*
```

If you keep it "as-is" just with uncomenting the `values.yaml` and add a filter option, for example :  

```
filter:
  match: "kube.*"
  kubeURL: "https://kubernetes.default.svc.cluster.local:443"
  mergeLog: "On"
  mergeLogKey: "data"
  keepLog: "On"
  k8sLoggingParser: "On"
  k8sLoggingExclude: "On"

extraFilters: |
  [FILTER]
      Use_Kubelet      On
```

You will get a `configmap` like this one, which is not working with fluentbit (it starts but won't take your extra options) :

```
    [FILTER]
        Name                kubernetes
        Match               kube.*
        Kube_URL            https://kubernetes.default.svc.cluster.local:443
        Merge_Log           On
        Merge_Log_Key       data
        Keep_Log            On
        K8S-Logging.Parser  On
        K8S-Logging.Exclude On
    [FILTER]
        Use_Kubelet            On
```

and the error in fluentbit logs:
```
Invalid configuration value at FILTER.Name
```

## Example 2:

Let's now try to use extra filter **without** the title, such as this example:

```
filter:
  match: "kube.*"
  kubeURL: "https://kubernetes.default.svc.cluster.local:443"
  mergeLog: "On"
  mergeLogKey: "data"
  keepLog: "On"
  k8sLoggingParser: "On"
  k8sLoggingExclude: "On"

extraFilter: |
      Use_Kubelet            On
```

you will get a `configmap` like this one, which is not working with fluentbit (it starts but won't take your extra options) :
```
    [FILTER]
        Name                kubernetes
        Match               kube.*
        Kube_URL            https://kubernetes.default.svc.cluster.local:443
        Merge_Log           On
        Merge_Log_Key       data
        Keep_Log            On
        K8S-Logging.Parser  On
        K8S-Logging.Exclude On
    Use_Kubelet            On
```

and the error in fluentbit logs:
```
[  Error] Error in line 24: Invalid indentation level
Error: Configuration file contains errors. Aborting
```

[...]

PR #658, opened 2021-12-13
Fix configmap generation when using extra option

ltupin:
### Issue

https://github.com/aws/eks-charts/issues/657

### Description of changes
When you want to use extra `[FILTER]` `[INPUT]` and `[OUTPUT]`  options it is confusing in comments to leave the title itself. See details and examples in https://github.com/aws/eks-charts/issues/657

That's why I want to propose a PR to indent from 4 to 8, any extra option and to remove the block configuration title `[FILTER]` `[INPUT]` and `[OUTPUT]` in comments in values file. 

The only block within I didn't try it is the PARSER so that's why I don't submit any change for this one (but maybe it is needed too).

### Checklist
- [X] Added/modified documentation as required (such as the `README.md` for modified charts)
- [X] Incremented the chart `version` in `Chart.yaml` for the modified chart(s)
- [X] Manually tested locally with helm template command and on EKS cluster
- [X] Make sure the title of the PR is a good description that can go into the release notes

### Testing

I apply it locally and it works, my `values.yaml` file:

```
filter: 
  match: "kube.*"
  kubeURL: "https://kubernetes.default.svc.cluster.local:443"
  mergeLog: "On"
  mergeLogKey: "data"
  keepLog: "On"
  k8sLoggingParser: "On"
  k8sLoggingExclude: "On"

extraFilters: |
   DNS_Retries         3
   DNS_Wait_Time       10
```

Command:
```
helm --debug template -s templates/configmap.yaml aws-for-fluent-bit  aws-for-fluent-bit -f values.yaml
```

Output:
```
    [FILTER]
        Name                kubernetes
        Match               kube.*
        Kube_URL            https://kubernetes.default.svc.cluster.local:443
        Merge_Log           On
        Merge_Log_Key       data
        Keep_Log            On
        K8S-Logging.Parser  On
        K8S-Logging.Exclude On
        DNS_Retries         3
        DNS_Wait_Time       10
```

**I deploy it on my EKS clusters and it works.**

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #693, opened 2022-02-15
fix(extra): handle existing and additional output, filter, extra

JBOClara:
### Issue

To fix https://github.com/aws/eks-charts/issues/657  this PR https://github.com/aws/eks-charts/pull/658 remove the ability to add new output, filter or input.

We need two ways to customize the configmap:
- add extra to existing config
- add additional config

### Checklist
- [X] Added/modified documentation as required (such as the `README.md` for modified charts)

### Breaking change

To rollback to the behavior previous to #658 but keep the flexibility to add extra parameters to existing input, filter ...
I've nested all extraXXX inside their block, for instance:

```
extraInputs: | 
   foo: bar
```

become

```
input:
  extraInputs: |
   foo: bar
```

And also, previous #658 usage: 

```
extraFilters: |
  [FILTER]
      Name modify
      Match *
      Set src_type kubernetes
```

became:

```
additionalFilters: |
  [FILTER]
      Name modify
      Match *
      Set src_type kubernetes
```

### Testing

I've tested elasticsearch.extraOutputs, serviceMonitor, service.extraService, extraFilters, extraInputs

### DIFF
--- a/stable/aws-for-fluent-bit/Chart.yaml
+++ b/stable/aws-for-fluent-bit/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-for-fluent-bit
 description: A Helm chart to deploy aws-for-fluent-bit project
-version: 0.1.13
+version: 0.1.14
 appVersion: 2.21.5
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
--- a/stable/aws-for-fluent-bit/README.md
+++ b/stable/aws-for-fluent-bit/README.md
@@ -42,7 +42,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
 | `extraInputs` | Append to input with this value | `""` |
 | `additionalInputs` | Adding more inputs with this value | `""` |
 | `filter.*` | Values for kubernetes filter | |
-| `extraFilters` | Append to filters with value |
+| `extraFilters` | Append to filter with value |
 | `additionalFilters` | Adding more filters with value |
 | `cloudWatch.enabled` | Whether this plugin should be enabled or not [details](https://github.com/aws/amazon-cloudwatch-logs-for-fluent-bit) | `true` | ✔
 | `cloudWatch.match` | The log filter | `*` | ✔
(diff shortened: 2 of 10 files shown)


PettitWesley (approved):

ltupin:
Hi there @JBOClara. I am confused and dont get why you mean my PR remove the ability to add new output, filter or input ?

stanvit:
#658  broke our configuration as we were using `extraFilters` and `extraOutputs` to add extra  filters and outputs. With `indent 8` the section names are not aligned properly and the config is not valid anymore.

Example:

```
extraFilters: |
  [FILTER]
      Name modify
      Match *
      Set src_type kubernetes
```

results in
```
[FILTER]
    Name                kubernetes
    Match               kube.*
    Kube_URL            https://kubernetes.default.svc.cluster.local:443
    Merge_Log           Off
    Merge_Log_Key       data
    Keep_Log            On
    K8S-Logging.Parser  On
    K8S-Logging.Exclude On
    [FILTER]
        Name modify
        Match *
        Set src_type kubernetes
```

ltupin:
@stanvit maybe Im wrong but in #658 I suggested to indent 8 **AND** to remove the block in the comment to avoid your situation (here [FILTER] in your example) if you just "remove" the second [FILTER] block, indent will be ok and your configuration will be ok.

Also, the thing I discover and try to explained in my issue is that if you keep two blocks, the configuration wont be take into account by fluentbit, did you double check it ?

So instead of doing this:

`extraFilters: |
  [FILTER]
      Name modify
      Match *
      Set src_type kubernetes`

doing that will fix your indentation problem:
`extraFilters: |
      Name modify
      Match *
      Set src_type kubernetes`

Please have a look to my explanations in #657 (Example 1 and Example 2). Several people where waiting for this fix (not only me) so I hope we wont rollback without better understanding of all problems we encounter and makes everyone happy :-D

JBOClara:
This PR is meant to keep the best of both worlds.

Be able to add parameters into existing output, filter and be able to add new output, filter ... 

I think this PR will fix the issue @stanvit mentioned, introduced by #658

ltupin:
Thanks a lot for your comment @JBOClara, I got it now !

We have two issue to fix:

1. if we do as I propose it breaks the ability to have several filters such as:
```
 [FILTER]
    Name                foo
    Match               kube.*

 [FILTER]
    Name                bar
    Match               *
```
2. if we do as you propose it breaks the ability to just add new options to the default existing filter !

We need to find a solution to fix both ! Thinking in progress...

JBOClara:
I concur with your first statement.

But I disagree with the second one :p 

With this very PR:

```
extraFilters: |
    optionadded    true
```

Render:
```
    [FILTER]
        Name                kubernetes
        Match               kube.*
        Kube_URL            https://kubernetes.default.svc.cluster.local:443
        Merge_Log           On
        Merge_Log_Key       data
        Keep_Log            Off
        K8S-Logging.Parser  On
        K8S-Logging.Exclude On
        optionadded    true
```

https://github.com/aws/eks-charts/blob/30623c6636ba95e4d260e7692919fda47ce8fb9d/stable/aws-for-fluent-bit/templates/configmap.yaml#L51

ltupin:
Awesome, we are agree now 👍 thx a lot @JBOClara !

So to add a config parameter to existing default filter use `extraFilters` without `[FILTER]` and to add a new filter use `additionalFilters` with `[FILTER]` block.

JBOClara:
@hossain-rayhan @PettitWesley @stanvit merging this PR will make everone happy :p

hossain-rayhan (approved):

PR #715, opened 2022-03-19
fix: set proper value name for additional parsers of aws-for-fluent-bit

witalisoft:
### Issue

n/a

### Description of changes

There is a skew between value names between templates and docs/values.yaml for additional parsers of `aws-for-fluent-bit`.

### Checklist
- [X] Added/modified documentation as required (such as the `README.md` for modified charts)
- [X] Incremented the chart `version` in `Chart.yaml` for the modified chart(s)
- [X] Manually tested. Describe what testing was done in the testing section below
- [X] Make sure the title of the PR is a good description that can go into the release notes

### Testing

not needed

### DIFF
--- a/stable/aws-for-fluent-bit/README.md
+++ b/stable/aws-for-fluent-bit/README.md
@@ -38,7 +38,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
 | `serviceAccount.create` | Whether a new service account should be created | `true` |
 | `service.extraService` | Append to existing service with this value | `""` |
 | `service.parsersFiles` | List of available parser files | `/fluent-bit/parsers/parsers.conf` |
-| `service.additionalParsers` | Adding more parsers with this value | `""` |
+| `service.extraParsers` | Adding more parsers with this value | `""` |
 | `input.*` | Values for Kubernetes input | |
 | `extraInputs` | Append to existing input with this value | `""` |
 | `additionalInputs` | Adding more inputs with this value | `""` |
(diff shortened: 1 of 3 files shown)


stable/aws-for-fluent-bit/README.md
```diff
@@ -38,7 +38,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
 | `serviceAccount.create` | Whether a new service account should be created | `true` |
 | `service.extraService` | Append to existing service with this value | `""` |
 | `service.parsersFiles` | List of available parser files | `/fluent-bit/parsers/parsers.conf` |
-| `service.additionalParsers` | Adding more parsers with this value | `""` |
+| `service.extraParsers` | Adding more parsers with this value | `""` |
```
PettitWesley, line 41: renaming the key seems like not a good idea, won't this break folks who depend on it??

PettitWesley: Sorry I didn't read the description- you're just updating the docs to reflect what the actual template supports right?

stable/aws-for-fluent-bit/values.yaml
```diff
@@ -20,7 +20,7 @@ service:
   #   HTTP_PORT    2020
   parsersFiles:
     - /fluent-bit/parsers/parsers.conf
-  # additionalParsers: |
+  # extraParsers: |
```
PettitWesley, line 23: hmmm it seems here we are changing the actual values in the template which may break folks?

PR #743, opened 2022-04-28
[aws-for-fluent-bit] User should be able to provide an existing fluent-bit configmap

AntonAleksandrov13:
### Description of changes

We are migrating from fluent-bit installation made based on  [EKS deployment guide](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-setup-logs-FluentBit.html). This article creates a set of custom fluent-bit configs which are not easy to migrate considering how fluent-bit populated now. I believe that a user should be able to provide own configmap with custom configurations.

### Checklist

### Testing

`helm upgrade --install aws-for-fluent-bit  ./` - creates a managed cm and mounts the cm

`helm upgrade --install aws-for-fluent-bit --set existingConfigMapName=my-custom-cm ./` - mounts an existing cm
 
### DIFF
--- a/stable/aws-for-fluent-bit/README.md
+++ b/stable/aws-for-fluent-bit/README.md
@@ -33,7 +33,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
 | `image.tag` | Image tag to deploy | `2.21.5`
 | `image.pullPolicy` | Pull policy for the image | `IfNotPresent` | ✔
 | `imagePullSecrets` | Docker registry pull secret | `[]` |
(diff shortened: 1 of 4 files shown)


PettitWesley:
@AntonAleksandrov13 If you provide your own config map, what's the value of using this chart? If you have your own config already set in a config map, I'd just setup my own daemonset with that config... why not do that?

AntonAleksandrov13:
@PettitWesley I do agree that providing completely custom config might be a bit too much, so the chart is losing it's value. However, the current setup is limiting and restrictive. For example adding lua functions for datetime conversion is not possible as far as I can tell.

 I can rework my PR to fit your structure a bit more and fit needed functionality. Would that work for you?

PR #778, opened 2022-06-22
Add support for configuring tags included with auto created log groups

choppedpork:
### Issue

n/a

### Description of changes

I've added the ability to set the `new_log_group_tags` option value for Fluent Bit Plugin for CloudWatch Logs. I'm currently starting work on migrating away from the FluentD CloudWatch implementation and this is the only bit of functionality I'm missing - it's already supported by the plugin itself, just not the chart.

### Checklist

### Testing

I've tested this change by deploying the helm chart to a test cluster, verifying the Fluent Bit logs and checking the tags on log groups in CloudWatch. Note that contrary to the [README of the plugin itself](https://github.com/aws/amazon-cloudwatch-logs-for-fluent-bit/blob/mainline/README.md?plain=1#L38) I had to skip the double quotes as otherwise I was getting the following error back from the API:

    Member must have length less than or equal to 128, Member must have length greater than or equal to 1, Member must satisfy regular expression pattern: ^([\\p{L}\\p{Z}\\p{N}_.:/=+\\-@]+)$

My test tag set didn't not include spaces however I've ran an additional test with spaces in the tag value and it's all worked great!

### DIFF
--- a/stable/aws-for-fluent-bit/README.md
+++ b/stable/aws-for-fluent-bit/README.md
@@ -56,6 +56,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
 | `cloudWatch.logFormat` | An optional parameter that can be used to tell CloudWatch the format of the data. A value of json/emf enables CloudWatch to extract custom metrics embedded in a JSON payload. See the Embedded Metric Format. |  |
 | `cloudWatch.roleArn` | ARN of an IAM role to assume (for cross account access). |  |
 | `cloudWatch.autoCreateGroup` | Automatically create the log group. Valid values are "true" or "false" (case insensitive). | true |
+| `cloudWatch.newLogGroupTags` | Tags to include with auto created log groups. Map of key/value pairs. | {} |
 | `cloudWatch.endpoint` | Specify a custom endpoint for the CloudWatch Logs API. |  |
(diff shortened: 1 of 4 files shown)


stable/aws-for-fluent-bit/README.md
```diff
@@ -56,6 +56,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
 | `cloudWatch.logFormat` | An optional parameter that can be used to tell CloudWatch the format of the data. A value of json/emf enables CloudWatch to extract custom metrics embedded in a JSON payload. See the Embedded Metric Format. |  |
 | `cloudWatch.roleArn` | ARN of an IAM role to assume (for cross account access). |  |
 | `cloudWatch.autoCreateGroup` | Automatically create the log group. Valid values are "true" or "false" (case insensitive). | true |
+| `cloudWatch.newLogGroupTags` | Tags to include with auto created log groups. Map of key/value pairs. | {} |
```
PettitWesley, line 59: Can you edit this slightly to make it even more clear- what you have right now is good, but let's say explicitly it only works if you set auto_create_group true and the log group did not already exist.

choppedpork: Thanks - that's now done!

PR #783, opened 2022-07-08
Add ability to override value of prometheus.io/scrape pod annotation

rwilson-release:
### Issue

Could not find an existing issue. Several clusters monitored with datadog agent have reported custom metrics like `rest_client_request_latency_seconds.bucket` and tags trace back to the installed AWS load balancer controller Prometheus metrics port.

### Description of changes

This patch allows values override for the annotation so that extra metrics will not be collected.

### Checklist

### Testing

1. Applied the chart without `--set prometheusScrape`: value of annotation is true
2. Applied the chart with `--set prometheusScrape=true`: value of annotation is true
3. Applied the chart with `--set prometheusScrape=false`: value of annotation is false


cep21:
Hi,

I would also like this change.  It's a big driver of our current datadog bill

cep21 (approved):

cep21 (approved):

kishorj:
@rwilson-release, these changes will get lost when we sync the chart from the upstream repo. Please create the PR in the upstream repo https://github.com/kubernetes-sigs/aws-load-balancer-controller/ instead.

Issue #856, opened 2022-12-07
`policy/v1beta1` has been deprecated since kubectl version 1.25


portega-inbrain:
The PodSecurityPolicy configuration has a dependency on `policy/v1beta1`
https://github.com/aws/eks-charts/blob/8e82f74d75221964d604d3c7b8c70da10160b88e/stable/aws-node-termination-handler/templates/psp.yaml#L2.
However this has been deprecated in more recent `kubectl` versions [PodDisruptionBudget](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125), [PodSecurityPolicy](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#psp-v125) in https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25.
This is causing the `aws-node-termination-handler` to fail during installation with message
```
Error: unable to build kubernetes objects from release manifest: resource mapping not found
for name: "aws-node-termination-handler" namespace: "" from "":
no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
ensure CRDs are installed first
```

Please, see the original issue leading to this https://github.com/dandi/dandi-hub/issues/52.

razorsk8jz:
This is also affecting myself when deploying aws-for-fluent-bit
https://github.com/aws/eks-charts/blob/69c1e0e46cc2ce2a4200535799ce99505e65f824/stable/aws-for-fluent-bit/templates/psp.yaml#L1

bakayolo:
Same issue here, just upgraded my cluster to 1.25 and I can't upgrade or even delete the node termination handler anymore

LeoSpyke:
Same here, even though the message is slightly different:
```
Error: unable to build kubernetes objects from release manifest: resource mapping not found
for name: "aws-for-fluent-bit" namespace: "" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
ensure CRDs are installed first
```

I'm actually trying to use it from the EKS Blueprints for Terraform by configuring the value `.rbac.pspEnabled` to `false` both via a new `values.yaml` and by manually configuring the 
```terraform
set {
    name  = "rbac.pspEnabled"
    value = false
}
```
but I'm getting the same error.

mvanbaak:
> Environment:
> 
> * EKS: 1.25
> * Chart version: 0.1.24
> * Application version: 2.28.4

[...]

Issue #810, opened 2022-08-21
aws-for-fluent-bit - remove PodSecurityPolicy


joebowbeer:
**Is your feature request related to a problem?**
There is no option to disable creation of [PodSecurityPolicy](https://kubernetes.io/docs/concepts/security/pod-security-policy/) (psp), which was deprecated in k8s 1.21 and will be removed in k8s 1.25

Relates to #476 

**Is your feature request related to a specific Helm chart, if yes mention name of the chart?**
aws-for-fluent-bit

**Describe the solution you'd like**
Add `podSecurityPolicy.create` with default value of `false`. 

See https://github.com/fluent/helm-charts/blob/main/charts/fluent-bit/values.yaml#L35

**Describe alternatives you've considered**
* kustomize
* https://fluent.github.io/helm-charts

yakir-shriker:
+1

yacov-nagose:
+1

joebowbeer:
Fixed by #888

PR #888, opened 2023-01-18
add ability to turn PodSecurityPolicy on or off in aws-for-fluent-bit

razorsk8jz:
### Issue

In kubernetes 1.25 pod security policies were deprecated - see issue https://github.com/aws/eks-charts/issues/856 and https://github.com/aws/eks-charts/issues/810

### Description of changes

Adding the ability to turn off the pod security policy will allow people to use the chart if they are on kubernetes 1.25+

### Checklist

### Testing

I was able to deploy the chart by setting rbac.pspEnabled=false

### DIFF
--- a/stable/aws-for-fluent-bit/templates/psp.yaml
+++ b/stable/aws-for-fluent-bit/templates/psp.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.rbac.pspEnabled }}
 apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
@@ -35,3 +36,4 @@ spec:
       - min: 1
         max: 65535
   readOnlyRootFilesystem: false
+{{- end }}
--- a/stable/aws-for-fluent-bit/Chart.yaml
+++ b/stable/aws-for-fluent-bit/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-for-fluent-bit
 description: A Helm chart to deploy aws-for-fluent-bit project
-version: 0.1.22
+version: 0.1.23
 appVersion: 2.21.5
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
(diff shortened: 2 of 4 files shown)


stable/aws-for-fluent-bit/templates/psp.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.rbac.pspEnabled }}
```
PettitWesley, line 1: @razorsk8jz are you sure that you want to wrap this entire file with this option? I think that means this existing file will not get generated unless you have this new default false field enabled. That seems non-desirable. 

Can you please explain the intention behind this? Should the option be default true?

I also see here that pod security policy is removed in newer k8s version: https://kubernetes.io/docs/concepts/security/pod-security-policy/

So I'm not sure it makes sense to move forward with this.

PettitWesley (changes requested):

razorsk8jz:
This also aligns with the request in https://github.com/aws/eks-charts/issues/810 to default to false

stable/aws-for-fluent-bit/templates/psp.yaml
PettitWesley: I see so the intention is to make it default false because k8s has removed pod security policy: https://github.com/aws/eks-charts/issues/810

PettitWesley: The chart seems to currently have this which defaults false, shouldn't this be used to gate creating this file or not?? https://github.com/fluent/helm-charts/blob/main/charts/fluent-bit/values.yaml#L35

PettitWesley: Sorry I'm really a novice at some of the things, I need you to explain it to me.

razorsk8jz:
Looking for updates if changes needed.

PettitWesley (changes requested):
@razorsk8jz I think you need to update the README with the new option before this can be merged: https://github.com/aws/eks-charts/blob/master/stable/aws-for-fluent-bit/README.md

razorsk8jz:
Readme Updated

stable/aws-for-fluent-bit/templates/psp.yaml
PettitWesley: I understand now, this makes it disabled by default, which we want since its deprecated.

PettitWesley (approved):
Tested, and it works, thanks!

```
helm template --dry-run . --set rbac.pspEnabled="true"
```

PR #899, opened 2023-02-14
Fixing missing prefix for extraInputs

perry-mitchell:
### Issue

_NA_.

### Description of changes

The `extraInputs` option is actually `input.extraInputs` according to the configmap template.

### Checklist

### Testing

### DIFF
--- a/stable/aws-for-fluent-bit/README.md
+++ b/stable/aws-for-fluent-bit/README.md
@@ -39,7 +39,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
 | `service.parsersFiles` | List of available parser files | `/fluent-bit/parsers/parsers.conf` |
 | `service.extraParsers` | Adding more parsers with this value | `""` |
 | `input.*` | Values for Kubernetes input | |
-| `extraInputs` | Append to existing input with this value | `""` |
+| `input.extraInputs` | Append to existing input with this value | `""` |
 | `additionalInputs` | Adding more inputs with this value | `""` |
 | `filter.*` | Values for kubernetes filter | |
 | `filter.extraFilters` | Append to existing filter with value |


stable/aws-for-fluent-bit/README.md
```diff
@@ -39,7 +39,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
 | `service.parsersFiles` | List of available parser files | `/fluent-bit/parsers/parsers.conf` |
 | `service.extraParsers` | Adding more parsers with this value | `""` |
 | `input.*` | Values for Kubernetes input | |
```
PettitWesley, line 41: nit: Can you fix the language above too? This is the kubernetes tail log file input.

stable/aws-for-fluent-bit/README.md
```diff
@@ -39,7 +39,7 @@ helm delete aws-for-fluent-bit --namespace kube-system
...
-| `extraInputs` | Append to existing input with this value | `""` |
+| `input.extraInputs` | Append to existing input with this value | `""` |
```
PettitWesley, line 42: I think what this is meant to do is to append to the kubernetes tail input. Because when I do a simple silly test:

```
$ helm template --dry-run . --set input.extraInputs="[INPUT]somekey someval"
...

  fluent-bit.conf: |
    [SERVICE]
        Parsers_File /fluent-bit/parsers/parsers.conf

    [INPUT]
        Name              tail
        Tag               kube.*
        Path              /var/log/containers/*.log
        DB                /var/log/flb_kube.db
        Parser            docker
        Docker_Mode       On
        Mem_Buf_Limit     5MB
        Skip_Long_Lines   On
        Refresh_Interval  10
        [INPUT]somekey someval
```

The new data is added indented to the existing tail input. So you could use this to set other values in it, like for example Ignore_Older:

```
$ helm template --dry-run . --set input.extraInputs="Ignore_Older 1d"
  fluent-bit.conf: |
    [SERVICE]
        Parsers_File /fluent-bit/parsers/parsers.conf

    [INPUT]
        Name              tail
        Tag               kube.*
        Path              /var/log/containers/*.log
        DB                /var/log/flb_kube.db
        Parser            docker
        Docker_Mode       On
        Mem_Buf_Limit     5MB
        Skip_Long_Lines   On
        Refresh_Interval  10
        Ignore_Older 1d
```

I think this was added because the `input.*` only allows you to customize a few things: https://github.com/aws/eks-charts/blob/master/stable/aws-for-fluent-bit/values.yaml#L29

So basically, the name here is wrong/misleading. It should be something like `input.extraSettings`.

I think the `additionalInputs` key is for adding other other inputs. 

I don't know how this works- can we bump the chart version and change the setting name?

If not, can you please make the help text here much more clear: "Append other config options to the kubernetes tail input" something like that.

perry-mitchell: Thanks @PettitWesley - While I understand your points, I simply wanted to correct a typo in the readme. The functionality and naming therein should be out of scope for this, right? At least people working with the current version will then not make the same mistake I did, as the name will be correct.

Imo all of the names need to be reworked, because they're not consistent and are all confusing:
 * service.extraParsers
 * input.*
 * extraInputs
 * additionalInputs
 * filter.*
 * filter.extraFilters
 * additionalFilters

PettitWesley: Fair point. I still think the language in your PR needs to be edited even just for this option. It should be clear what "existing input" means. 

> Append to existing kubernetes log file tail input with this value

would be prefered by me

PettitWesley: there is also another issue here that the CI only lets you merge updates that change the chart version, which I'll have to look into.

PR #913, opened 2023-02-28
Fix Prometheus PVC pending problem

ysdongAmazon:
### Issue

Fix Prometheus PVC pending problem

### Checklist
- [1] Added/modified documentation as required (such as the `README.md` for modified charts)
- [1] Incremented the chart `version` in `Chart.yaml` for the modified chart(s)
- [1] Manually tested. Describe what testing was done in the testing section below
- [1] Make sure the title of the PR is a good description that can go into the release notes

### Testing

### DIFF
--- a/stable/appmesh-prometheus/README.md
+++ b/stable/appmesh-prometheus/README.md
@@ -21,7 +21,35 @@ helm upgrade -i appmesh-prometheus eks/appmesh-prometheus \
 --namespace appmesh-system
 ```
 
-For Prometheus persistent storage you can create a [PersistentVolumeClaim](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)
+Optional: Persist your data  
+If you do not persist your Prometheus data, then it will only exist as long as the Prometheus pod is running. For Prometheus persistent storage, you need to use PersistentVolumeClaim.  
+As for the volume plugin, we use EBS CSI Driver as an example, but you can use other popular volume plugins like NFS, Ceph etc.  
+In configuration, replace *your-appmesh-cluster* as your EKS appmesh cluster name and *your-aws-account* as your AWS account ID.
+
+Enable EBS CSI Driver:  
+- Initialize iam-oidc-provider
+```
+eksctl utils associate-iam-oidc-provider --region=us-west-2 --cluster=your-appmesh-cluster --approve
+```
+- Create IAM role
+```
+eksctl create iamserviceaccount \
+  --name ebs-csi-controller-sa \
+  --namespace kube-system \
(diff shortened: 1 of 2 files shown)


EdwardXF (approved):

stable/appmesh-prometheus/README.md
```diff
@@ -21,6 +21,29 @@ helm upgrade -i appmesh-prometheus eks/appmesh-prometheus \
...
+- Initialize iam-oidc-provider
+```
+eksctl utils associate-iam-oidc-provider --region=us-west-2 --cluster=appmeshtest --approve
+```
+- Create IAM role
+```
+eksctl create iamserviceaccount \
+  --name ebs-csi-controller-sa \
+  --namespace kube-system \
+  --cluster appmeshtest \
```
joesbigidea, line 34: The customer would need to replace cluster name with their cluster name right?

stable/appmesh-prometheus/README.md
```diff
@@ -21,6 +21,29 @@ helm upgrade -i appmesh-prometheus eks/appmesh-prometheus \
 --namespace appmesh-system
 ```
 
+Enable EBS CSI Driver:
```
joesbigidea, line 24: I see in the chart that use of a persistent volume is controlled by persistentVolumeClaim.claimName. Is this section only needed if that is configured?

joesbigidea, line 24: Also I wonder if this belongs in the prerequisites section. Maybe something about "To use a persistent volume for storing data you'll need the EBS CSI Driver installed. This isn't necessarily part of installing this chart, since it's one time setup for the cluster.

ysdongAmazon: I think so. Because the original README creates a PersistentVolumeClaim, the PVC is pending without the EBS driver, but there may be other ways to mount it, such as NFS.

ysdongAmazon: Thank you for the notificaiton, Joe. I just made some research, PVC is not mandatory for Prometheus to work, but highly recommended. Without a PVC, Prometheus data will be stored on the local filesystem of the container in which it is running. If the container is restarted or recreated, the data will be lost. This is not ideal for a time-series database that needs to store data over a long period of time. Because the original text already emphasizes the need to use persistent volume, I will add the Optional label to this section for further notice.

stable/appmesh-prometheus/README.md
ysdongAmazon: Thank you, I re-corrected with a replaceable label here for cluster name

joesbigidea (approved):
Looks great!

Issue #1039, opened 2023-12-13
[aws-for-fluent-bit] upgrade version used by Helm chart


paulfouquet:
We are experiencing [a bug](https://github.com/aws/aws-for-fluent-bit/issues/354) in [the current version of `aws-for-fluent-bit` used by the helm chart ](https://github.com/aws/eks-charts/blob/2ca1cb467a7ded4a608ab314b5cecf04dc7d758c/stable/aws-for-fluent-bit/Chart.yaml#L5)

This bug [should be fixed](https://github.com/aws/aws-for-fluent-bit/issues/526) [in the version 2.32 of `aws-for-fluent-bit`](https://github.com/aws/aws-for-fluent-bit/releases/tag/v2.32.0)

The `aws-for-fluent-bit` Helm Chart is currently using version 2.31.12.20231011 of `aws-for-fluent-bit` and should upgrade to version 2.32.

I am willing to manually change the version used in the Helm Chart, but don't know what the consequences are (are they new parameters/configuration required by this new version that are not taking into account by the current version of the Helm?).

crdrost:
I was going to add this to my "lots of little upgrades" PR but when I investigated I noticed that the requested version has not yet been marked as "stable" in aws-for-fluent-bit and probably needs to be before this repo's `stable/` charts are updated. 

I don't work at Amazon but they are probably gearing up for a release with a desire to not mess things up around Christmas when they might be short-staffed and also trying to finish other tasks frantically before the quarter is over... I would maybe therefore guess that there's a lot on their collective plates and they'll be updating it in early January?

paulfouquet:
Thanks for the update @crdrost!

michemache:
`aws-for-fluent-bit` v2.32 is now stable: https://github.com/aws/aws-for-fluent-bit/blob/mainline/AWS_FOR_FLUENT_BIT_STABLE_VERSION

Are there any plans to update the Helm chart soon?

bendu:
Hi there

I am one of the new maintainers of AWS for Fluent Bit. I'll try to look into this, but it may take a little time for me to study the commit history to see how to get the version updated and released. (If you are familiar with the necessary changes needed, please feel free to send a PR)

Feel free to reach out to me after about 3 weeks if you don't see any update.

cc: @PettitWesley

andrewhharmon:
[...]

PR #1373, opened 2026-09-17
Add workflow to auto-bump aws-for-fluent-bit chart

gian25-work:
Adds a weekly GitHub Actions workflow that checks the latest release of [aws/aws-for-fluent-bit](https://github.com/aws/aws-for-fluent-bit/releases) and opens a PR bumping `appVersion`, `image.tag` and the chart version. Mirrors the existing `update-eks-chart.yaml` (cron + `hack/` script + `peter-evans/create-pull-request`), but needs no AWS credentials since it only reads the public GitHub releases API.

## Version handling
- **Latest is picked by semver** (`sort -V`), not by release date. GitHub lists releases newest-first by creation date, so a backport to an older line (e.g. a 2.x CVE rebuild published after 3.4.16) would otherwise be picked and downgrade the chart. This also means 2.x releases are naturally ignored while 3.x exists, and a future 4.x is picked up instead of silently ignored.
- **Current ahead of latest** (semver compare) is a no-op, so a manual bump to a newer version is never reverted.
- **Chart version bump follows the size of the image jump**: a major image bump (e.g. 2.x → 3.x) increments the chart minor version and resets patch, as done manually in #1294; minor/patch/build image bumps increment the chart patch version.

Background: the chart is currently bumped by hand (last on 2026-02-03, #1294) and lags the image by several releases. See #1039 and #1372.

## Testing
Ran `./hack/update-aws-for-fluent-bit.sh` locally against master:
- 3.2.1 → 3.4.16: only three lines change (Chart.yaml `version` 0.2.0 → 0.2.1, `appVersion`, values.yaml `image.tag`), `helm lint` passes, second run is a no-op.
- Simulated `appVersion: 2.34.3.20260910` → 3.4.16: chart version 0.2.0 → 0.3.0.
- Simulated `appVersion: 9.0.0`: no-op.

`image.tag` is updated with `sed` rather than `yq -i` because yq rewrites the block scalars and comment indentation in values.yaml.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gian25-work:
The logic of the script will need to be reviewed when the chart version is stable v1 instead of v0

